PR #2, opened 2018-05-23
Create basic modules and build files. 

melowe:
Very simple ServiceLocator for Spring configured beans..

### DIFF
--- a/nexus-app/src/main/java/com/github/nexus/api/SomeOtherObject.java
+++ b/nexus-app/src/main/java/com/github/nexus/api/SomeOtherObject.java
@@ -1,0 +1,21 @@
+
+package com.github.nexus.api;
+
+import javax.validation.constraints.NotNull;
+
+
+
+public class SomeOtherObject {
+ 
+    @NotNull
+    private String value;
+
+    public String getValue() {
+        return value;
+    }
+
+    public void setValue(String value) {
+        this.value = value;
+    }
+
+}
--- a/nexus-app/src/test/java/com/github/nexus/app/PublicApiIT.java
+++ b/nexus-app/src/test/java/com/github/nexus/app/PublicApiIT.java
@@ -44,7 +44,7 @@ public class PublicApiIT {
 
         javax.ws.rs.core.Response response = client
                 .target(SERVICE_URI)
-                .path("/api/version")
+                .path("/version")
                 .request()
                 .get();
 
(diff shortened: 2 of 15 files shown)


nexus-app/src/main/java/com/github/nexus/api/SomeOtherObject.java
```diff
@@ -0,0 +1,21 @@
+
+package com.github.nexus.api;
+
+import javax.validation.constraints.NotNull;
+
+
```
SatpalSandhu61, line 6: I know this is a very trivial point, but do we need so much whitespace?

nexus-app/src/test/java/com/github/nexus/app/PublicApiIT.java
```diff
@@ -44,7 +44,7 @@ public void requestVersion() {
 
         javax.ws.rs.core.Response response = client
                 .target(SERVICE_URI)
-                .path("/api/version")
```
SatpalSandhu61, line 47: Thanks - I thought I'd changed that, but obviously missed it!

service-locator/pom.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.github.nexus</groupId>
+        <artifactId>nexus</artifactId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>service-locator</artifactId>
```
SatpalSandhu61, line 9: I put the artifact details first in all the other poms, to make them consistent. It would be good to keep them all consistent, it just makes it easier when looking through all the poms or comparing them.

SatpalSandhu61 (approved):

nexus-app/src/main/java/com/github/nexus/api/SomeOtherObject.java
melowe: not trivial.. we can add that to the check style stuff

melowe:
Basic rest server and implementation created via service loader mechanism. Sample load to demonstrate usage.

PR #3, opened 2018-05-24
Added a basic TravisCI configuration file for Java & Maven

prd-fox:
The Travis config file runs the following commands:

mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
mvn test -B

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,1 @@
+language: java


melowe:
maven.javadoc.skip=true  .. I think if people are using javadocs then the build needs to validate that they are correct. executing the java doc plugin as part of the normal build will be enough

melowe (approved):

SatpalSandhu61 (approved):

prd-fox:
@melowe Yes you are correct. This is just the most basic build file and that is the default Travis gives. I will be adding more steps and in that I won't skip the java doc validation.

PR #6, opened 2018-05-25
Add more api endpoints and relevant object models

namtruong:
Add relevant model objects for requests and responses.
Add transaction resource and api endpoints
Add service layer with an empty implementation and tests

### DIFF
--- a/nexus-app/src/main/java/com/github/nexus/api/TransactionResource.java
+++ b/nexus-app/src/main/java/com/github/nexus/api/TransactionResource.java
@@ -1,0 +1,46 @@
+package com.github.nexus.api;
+
+import com.github.nexus.api.model.ReceiveRequest;
+import com.github.nexus.api.model.SendRequest;
+import com.github.nexus.service.TransactionService;
+
+import javax.validation.Valid;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Path("/transaction")
+public class TransactionResource {
+
+    private static final Logger LOGGER = Logger.getLogger(TransactionResource.class.getName());
+
+    private TransactionService transactionService;
+
+    public TransactionResource(TransactionService transactionService) {
+        this.transactionService = transactionService;
+    }
+
+    @POST
+    @Path("/send")
+    @Consumes({MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML})
+    public Response send(@Valid final SendRequest sendRequest){
+        LOGGER.log(Level.INFO,"POST send");
+        transactionService.send();
+        return Response.status(Response.Status.CREATED).build();
+    }
+
+    @POST
+    @Path("/receive")
+    @Consumes({MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML})
+    public Response receive(@Valid final ReceiveRequest receiveRequest){
+        LOGGER.log(Level.INFO,"POST receive");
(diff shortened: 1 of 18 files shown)


nexus-app/src/main/java/com/github/nexus/api/model/DeleteRequest.java
```diff
@@ -0,0 +1,17 @@
+package com.github.nexus.api.model;
+
+import javax.validation.constraints.NotNull;
+
+public class DeleteRequest {
```
SatpalSandhu61, line 5: Can you add java docs to the models please indicating what each is for (just put a TODO for any where you're not sure of the usage - just to make sure we populate them later)

nexus-app/src/main/java/com/github/nexus/api/TransactionResource.java
```diff
@@ -0,0 +1,46 @@
...
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Path("/transaction")
+public class TransactionResource {
```
SatpalSandhu61, line 17: To follow convention, should these be called Controllers rather than Resources?

melowe: They are view controllers can’t we not call them thus.. the controller idiom is an overhang from spring MVC .. Resource is the correct convention

SatpalSandhu61 (approved):

nexus-app/src/main/java/com/github/nexus/api/model/DeleteRequest.java
namtruong: yup sure

nexus-app/src/main/java/com/github/nexus/api/TransactionResource.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(TransactionResource.class.getName());
+
+    private TransactionService transactionService;
+
+    public TransactionResource(TransactionService transactionService) {
+        this.transactionService = transactionService;
```
prd-fox, line 24: Should we do a null check here for injected dependencies?

nexus-app/src/test/java/com/github/nexus/api/TransactionResourceTest.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+public class TransactionResourceTest extends JerseyTest {
+
+    @Mock
```
prd-fox, line 24: Can we avoid the use of these annotations? We should create our dependencies and test classes explicitly.

nexus-app/src/main/java/com/github/nexus/api/TransactionResource.java
namtruong: yes. this is skeleton. will add

prd-fox (changes requested):
See comment on test mocks.

nexus-app/src/test/java/com/github/nexus/api/TransactionResourceTest.java
namtruong: this is just preference - it can be changed to return new ResourceConfig()
                .register(new TransactionResource(transactionService));
I won't be bothered too much at this point

prd-fox (approved):

PR #9, opened 2018-05-30
Add more endpoints and tests

namtruong:
Add unit tests for the endpoints, and refactor example packages

### DIFF
--- a/nexus-app/src/main/java/com/github/nexus/api/TransactionResource.java
+++ b/nexus-app/src/main/java/com/github/nexus/api/TransactionResource.java
@@ -1,6 +1,8 @@
 package com.github.nexus.api;
 
+import com.github.nexus.api.model.DeleteRequest;
 import com.github.nexus.api.model.ReceiveRequest;
+import com.github.nexus.api.model.ResendRequest;
 import com.github.nexus.api.model.SendRequest;
 import com.github.nexus.service.TransactionService;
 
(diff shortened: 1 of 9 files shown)


nexus-app/src/main/java/com/github/nexus/api/TransactionResource.java
```diff
@@ -33,6 +44,16 @@ public Response send(@Valid final SendRequest sendRequest){
         return Response.status(Response.Status.CREATED).build();
     }
 
+    @POST
+    @Path("/sendraw")
+    public Response sendRaw(@Context final HttpHeaders headers, InputStream inputStream) throws IOException {
```
prd-fox, line 49: InputStream should be final

nexus-app/src/test/java/com/github/nexus/service/TransactionServiceTest.java
```diff
@@ -15,4 +15,19 @@ public void testSend(){
     public void testReceive(){
         transactionService.receive();
     }
+
+    @Test
+    public void testDelete(){
+        transactionService.delete();
```
prd-fox, line 21: Add some verification that the method was actually called?

melowe (approved):

nexus-app/src/test/java/com/github/nexus/service/TransactionServiceTest.java
namtruong: the verification should be at the upper level. this test should test what the method does - which is nothing at the mo

prd-fox: Oh right yes. Oops.

PR #10, opened 2018-05-30
Minimal intal jpa config and enure that implemention is 2.2 complient

melowe:


namtruong (approved):
can templatize the values at some point later

PR #11, opened 2018-05-30
Spring + JPA config

melowe:


namtruong (approved):

nexus-app/src/main/resources/META-INF/persistence.xml
```diff
@@ -4,10 +4,10 @@
     <provider>org.eclipse.persistence.jpa.PersistenceProvider</provider>
     <exclude-unlisted-classes>false</exclude-unlisted-classes>
     <properties>
-      <property name="javax.persistence.jdbc.url" value="jdbc:h2:mem:hello"/>
+<!--      <property name="javax.persistence.jdbc.url" value="jdbc:h2:mem:hello"/>
```
SatpalSandhu61, line 7: Should remove it from here if it's defined in nexus-spring.xml and no longer required here.

PR #12, opened 2018-05-30
KIll dead sample code including poc with xsd to generate model. Add j…

melowe:
…avax.transaction for annotataions to me used with spring tx.


prd-fox (approved):

melowe:
Use declarative transactions using spring's support javax.transaction annotation support and remove sample objects.

PR #20, opened 2018-06-01
Use h2 db file rather than in memory db due to sequence generation issues

melowe:
play well with jpa sequence generation.


namtruong (approved):

melowe:
Using h2 in memory causes some strange side effects with sequence generation. Changed code to use file based db.

PR #55, opened 2018-06-08
Add application configuration that can be used to configure services

prd-fox:
Add an initial version of configuration and its holder classes that can be sent around the application.

### DIFF
--- a/nexus-app/src/main/java/com/github/nexus/config/ConfigHolder.java
+++ b/nexus-app/src/main/java/com/github/nexus/config/ConfigHolder.java
@@ -1,0 +1,34 @@
+package com.github.nexus.config;
+
+import java.util.Objects;
+
+public enum ConfigHolder {
+
+    INSTANCE;
+
+    private Configuration config;
+
+    private boolean initialised;
+
+    ConfigHolder() {
+        this.initialised = false;
+    }
+
+    public synchronized void setConfiguration(final Configuration config) {
+        if(initialised) {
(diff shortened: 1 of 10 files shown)


namtruong (dismissed):

namtruong (approved):

nexus-app/src/main/java/com/github/nexus/config/ConfigHolder.java
```diff
@@ -0,0 +1,34 @@
+package com.github.nexus.config;
+
+import java.util.Objects;
+
+public enum ConfigHolder {
+
+    INSTANCE;
```
SatpalSandhu61, line 7: Could we not use longhand here, to make it clearer?
public static final MySingleton INSTANCE = new MySingleton();

prd-fox: Yeah, this was done to keep it outside the spring xml file, but then it somehow made its way in anyway. Definitely a refactoring consideration, but for now this works.

melowe: Using enum classes for singletons is pretty uncontroversial.

SatpalSandhu61: @Mark, sure but it's good to keep things simple. I'd never seen this before and I'm sure there are plenty of folks who'd have to do a quick google to work it out.

PR #62, opened 2018-06-09
Added JNaCL, a pure Java implementation of the NaCL standard.

prd-fox:
This can be used as an alternative to Kalium.

Although JNaCL has no need for native bindings (unlike Kalium), it
likely has poorer performance.

Benchmarking needs to be performed to see how much quicker one is
over the other.

### DIFF
--- a/nexus-encryption-jnacl/src/main/java/com/github/nexus/nacl/jnacl/Jnacl.java
+++ b/nexus-encryption-jnacl/src/main/java/com/github/nexus/nacl/jnacl/Jnacl.java
@@ -1,0 +1,251 @@
+package com.github.nexus.nacl.jnacl;
+
+import com.github.nexus.nacl.*;
+import com.neilalexander.jnacl.NaCl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.security.SecureRandom;
+import java.util.Arrays;
+import java.util.Objects;
+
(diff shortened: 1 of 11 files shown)


melowe (approved):

nexus-encryption-jnacl/src/main/java/com/github/nexus/nacl/jnacl/Jnacl.java
```diff
@@ -0,0 +1,251 @@
...
+import com.neilalexander.jnacl.NaCl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.security.SecureRandom;
+import java.util.Arrays;
+import java.util.Objects;
+
+import static com.neilalexander.jnacl.crypto.curve25519xsalsa20poly1305.*;
+
```
SatpalSandhu61, line 13: Can we start adding java docs as we go along please. We're building up quite a bit of tech debt otherwise.

nexus-encryption-jnacl/src/main/java/com/github/nexus/nacl/jnacl/SecretBox.java
```diff
@@ -0,0 +1,13 @@
+package com.github.nexus.nacl.jnacl;
+
+public interface SecretBox {
```
SatpalSandhu61, line 3: This *definitely* needs java docs!

melowe:
The code reads well for me. What will java docs achieve?

PR #90, opened 2018-06-15
Private key encryption

prd-fox:
Add a encryption/decryption class for private keys that mediates between Argon2 and NaCL

Add a model object to help marshall to/from the expect JSON structure

Add passwords to the configuration that can be used to decrypt keys
Change return types of properties to allow easier management of config in main app code
Add a transient config property that is made up of other key properties

### DIFF
--- a/nexus-app/src/main/java/com/github/nexus/key/EncryptedPrivateKey.java
+++ b/nexus-app/src/main/java/com/github/nexus/key/EncryptedPrivateKey.java
@@ -5,6 +5,10 @@ import com.github.nexus.argon2.ArgonOptions;
 import javax.json.Json;
 import javax.json.JsonObject;
 
+/**
+ * POJO encapsulating the structure of the "data" field in a private key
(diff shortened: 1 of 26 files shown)


nexus-app/src/main/java/com/github/nexus/key/EncryptedPrivateKey.java
```diff
@@ -0,0 +1,95 @@
+package com.github.nexus.key;
+
+import com.github.nexus.argon2.ArgonOptions;
+
+import javax.json.Json;
+import javax.json.JsonObject;
+
+/**
+ * POJO encapsulating the structure of the "data" field in a private key
+ * when it as been encrypted using Argon2/NaCL
```
SatpalSandhu61, line 10: typo 'as been' (or is it or northern speak...)

prd-fox: 'as bin

namtruong (approved):

PR #141, opened 2018-06-21
Download junixsocket and provide native lib directory as a start

melowe:
to integrating geth and nexus during integration testing.


prd-fox (approved):

nexus-test/pom.xml
```diff
@@ -35,12 +35,43 @@
                 <version>3.1.1</version>
                 <executions>
                     <execution>
+                        <id>inject-depednencies-as-properties</id>
```
SatpalSandhu61, line 38: Typo - needs correction

PR #143, opened 2018-06-21
Remove the 1 second delay when establishing the HTTP connection

prd-fox:


melowe (approved):

nexus-socket/src/main/java/com/github/nexus/socket/SocketServer.java
```diff
@@ -109,12 +107,6 @@ private boolean createHttpServerConnection() {
         while (!connected) {
             LOGGER.info("Attempting connection to HTTP server...");
             connected = httpProxy.connect();
-            try {
-                threadDelegate.sleep(1000);
```
SatpalSandhu61, line 113: Removal of the sleep will cause it to spin until it gets a connection. That'll just slow down the startup of the HTTP server (and cause it to generate a lot of log output).

prd-fox: This connection isn't made at the startup of the HTTP server, so I am not sure what you mean. Also, since it fetches the input and output stream before returning from the method, it will block until it gets a connection (or errors).

PR #146, opened 2018-06-24
If request for partyinfo fails, then must not try to parse the response

SatpalSandhu61:
If request for partyinfo fails, then must not try to parse the response (it was parsing the response and getting garbage buffer length).


nexus-app/src/main/java/com/github/nexus/node/PostDelegate.java
```diff
@@ -22,6 +26,10 @@ public PostDelegate(final Client client) {
             .request()
             .post(Entity.entity(data, MediaType.APPLICATION_OCTET_STREAM_TYPE));
 
+        if (Response.Status.OK.getStatusCode() != response.getStatus()) {
```
prd-fox, line 29: We should note where we are using these, as when we change endpoints to return more specific status codes, this will break

prd-fox (approved):

nexus-app/src/main/java/com/github/nexus/node/PostDelegate.java
```diff
@@ -22,6 +26,10 @@ public PostDelegate(final Client client) {
...
+            return null;
```
namtruong, line 30: can we filter out the response status which is OK in the poller, rather than return null here?

SatpalSandhu61: If I do that then we have to return the response object to the poller, so it'll leak details which I thought are better kept hidden in the PostDelegate.

PR #147, opened 2018-06-24
Socket server threading

prd-fox:
Separate the executor that accepts socket connections from the server file into its own class

### DIFF
--- a/nexus-socket/src/main/java/com/github/nexus/socket/HttpProxy.java
+++ b/nexus-socket/src/main/java/com/github/nexus/socket/HttpProxy.java
@@ -23,10 +23,6 @@ public class HttpProxy {
 
     private Socket socket;
 
-    private OutputStream os;
-
-    private InputStream is;
-
     private final SocketFactory socketFactory;
 
     /**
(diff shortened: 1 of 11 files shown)


nexus-socket/src/main/java/com/github/nexus/socket/HttpProxy.java
```diff
@@ -75,9 +65,12 @@ public void disconnect() {
     /**
      * Write data to the http connection.
      */
-    public void sendRequest(byte[] data) {
+    public void sendRequest(final byte[] data) {
         LOGGER.info("Sending HTTP request: {}", data);
+
         try {
+            final OutputStream os = socket.getOutputStream();
```
SatpalSandhu61, line 72: Is there any particular reason for creating the input/output stream per request, rather than once on construction?
It seems a bit inefficient.

prd-fox: It is probably less efficient that reusing the same stream.

However, the current behaviour creates a new HTTP connection for every call over the socket, so nothing has currently been lost.

Additionally, this is gearing up to allowing multiple calls at the same time, where we won't be able to use the same stream in case of a clash (which will almost certainly happen).

SatpalSandhu61 (approved):

PR #167, opened 2018-06-28
Add path config for keys

prd-fox:
Add reading the path from file on JAXB instantiated classes (that bypass the constructor)


melowe (approved):

nexus-config/src/main/java/com/github/nexus/config/LegacyPrivateKeyFileAdapter.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.json.Json;
+import javax.json.JsonObject;
+import javax.xml.bind.annotation.adapters.XmlAdapter;
+import java.io.IOException;
+import java.io.StringReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+
```
SatpalSandhu61, line 13: Can we have a javadoc pls explaining what the legacy file is and why we need an adapter

SatpalSandhu61 (approved):

nexus-config/src/main/java/com/github/nexus/config/LegacyPrivateKeyFileAdapter.java
prd-fox: Yes

PR #168, opened 2018-06-28
Documentation

melowe:


README.md
```diff
@@ -2,12 +2,72 @@
 A stateless JAVA application responsible for encryption and decryption of private transaction data and for off-chain private messaging.It is also responsible for generating and managing private key locally in each node in Quorum Network.
 
 ##Running Nexus
-Usage:
 
 `java -jar nexus-app/target/nexus-app-${version}-app.jar -c config.json`
 
+
+Probably best to copy the jar somewhere and create an alias
+<pre>
```
SatpalSandhu61, line 10: Can we avoid using html tags in the readme.md as it's not guaranteed for them to be supported in markdown (I replaced all the html tags previously to use standard markdown)

SatpalSandhu61 (changes requested):
Can you replace the html tags with markdown

SatpalSandhu61 (approved):

README.md
melowe: if you get get preformatted to work then sure

PR #226, opened 2018-07-07
Feature/dependency management packaging

melowe:


namtruong (approved):

argon2/pom.xml
```diff
@@ -16,19 +16,6 @@
         <version>1.0-SNAPSHOT</version>
     </parent>
 
-    <dependencyManagement>
-
-        <dependencies>
-
-            <dependency>
-                <groupId>de.mkammerer</groupId>
```
SatpalSandhu61, line 24: Shouldn't this dependency remain in the argon2 module as it's presumably only used there?

melowe: it needs rescoping as runtime so people who don't know any better (whoever they may be) don't start using it as a compile time dependency

PR #245, opened 2018-07-10
Refactoring some child modules

namtruong:
Refactoring some child modules


app/pom.xml
```diff
@@ -4,7 +4,7 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <artifactId>application</artifactId>
+    <artifactId>app</artifactId>
```
melowe: shouldn't this be tessera-app?

melowe (changes requested):
the artifactid and directory should be tessera-app. The final name does generate the file , but nexus (maven stuff) uses the artififact id regardless of the final name

melowe (dismissed):

melowe (dismissed):

melowe (approved):

PR #266, opened 2018-07-23
Make resend requests for all new parties to be discovered

prd-fox:
For any new party that is discovered via the node polling, make an attempt to resend all transactions they might have for all of the currently managed public keys.


tessera-app/src/main/java/com/github/tessera/sync/model/SyncableParty.java
```diff
@@ -0,0 +1,24 @@
+package com.github.tessera.sync.model;
+
+import com.github.tessera.node.model.Party;
+
+public class SyncableParty {
```
SatpalSandhu61, line 5: Could do with a javadoc here

tessera-app/src/main/java/com/github/tessera/node/PartyInfoPoller.java
```diff
@@ -3,11 +3,13 @@
...
 import com.github.tessera.node.model.PartyInfo;
+import com.github.tessera.sync.ResendPartyStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.ConnectException;
 import java.util.Objects;
+import java.util.Set;
 
 public class PartyInfoPoller implements Runnable {
```
SatpalSandhu61, line 14: Could you add a javadoc for this class whilst you're making changes pls?

SatpalSandhu61 (approved):

tessera-app/src/main/java/com/github/tessera/node/PartyInfoPoller.java
melowe: What isn't clear? I can see some java docs on interfaces might be useful. I really don't want the patch code with comments culture starting here..

tessera-app/src/main/java/com/github/tessera/sync/model/SyncableParty.java
prd-fox: Yes will do

tessera-app/src/main/java/com/github/tessera/node/PartyInfoPoller.java
prd-fox: Yes will do

SatpalSandhu61: @melowe Mark, I know there are different schools of thought on this; however we've always tried to follow the practice of ensuring classes/code have a brief javadoc which gives the *purpose*.
A class name can't give you this (unless you want class names to be a sentence).
It only takes a few seconds effort and it's a huge help to new members of the team and any folks who aren't familiar with the code (which will be a lot of people since this will be open source).

PR #274, opened 2018-07-24
Jdbc driver class name mapping

namtruong:
Use relevant driver class name based on jdbc url


config/src/main/java/com/github/tessera/config/util/JdbcDriverClassName.java
```diff
@@ -0,0 +1,35 @@
+package com.github.tessera.config.util;
+
+public enum JdbcDriverClassName {
```
SatpalSandhu61: Could you add a comment that this basically provides a mapping from a jdbc url to a driver class name

SatpalSandhu61 (dismissed):

prd-fox (approved):

PR #283, opened 2018-07-26
Add tls cert and key file paths and legacy config support

melowe:


prd-fox (approved):

config-cli/src/main/java/com/quorum/tessera/config/cli/LegacyCliAdapter.java
```diff
@@ -56,8 +56,7 @@ public CliResult execute(String... args) throws Exception {
                 .map(ConfigBuilder::from)
                 .orElse(ConfigBuilder.create());
 
-        
-        ConfigBuilder adustedConfig = applyOverrides(line,configBuilder);
+        ConfigBuilder adustedConfig = applyOverrides(line, configBuilder);
```
SatpalSandhu61, line 59: Typo - presumably that should be 'adjustedConfig'

PR #313, opened 2018-07-27
Slim data migration jar down and create shaded version to be used.

melowe:
Use log opt args.. Exclude Main class from coverage but port all the logic
ito executr object that returns exit value. Print exeception.getMessage
to cli user in the case of an error.

### DIFF
--- a/data-migration/src/main/java/com/quorum/tessera/data/migration/CmdLineExecutor.java
+++ b/data-migration/src/main/java/com/quorum/tessera/data/migration/CmdLineExecutor.java
@@ -1,0 +1,101 @@
+
+package com.quorum.tessera.data.migration;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Map;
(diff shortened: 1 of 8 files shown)


namtruong (approved):

data-migration/src/main/java/com/quorum/tessera/data/migration/CmdLineExecutor.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Map;
+import java.util.Objects;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+
+
```
SatpalSandhu61, line 16: A javadoc on this would help explain what it's for...

SatpalSandhu61 (approved):

PR #327, opened 2018-07-30
Reduce the time that the unix socket poller waits before listening for the next connection

prd-fox:


Krish1979 (approved):

tessera-app/src/main/java/com/quorum/tessera/threading/TesseraScheduledExecutor.java
```diff
@@ -50,7 +50,7 @@ public void start() {
             }
         };
 
-        this.executor.scheduleWithFixedDelay(exceptionSafeRunnable, rateInSeconds, rateInSeconds, TimeUnit.SECONDS);
+        this.executor.scheduleWithFixedDelay(exceptionSafeRunnable, rateInSeconds, rateInSeconds, TimeUnit.MILLISECONDS);
```
SatpalSandhu61, line 53: To avoid confusion, need to rename the local variable (and parameter passed into the constructor), from "rateInSeconds".

PR #337, opened 2018-07-31
Initial util for using reflection to generate cmd override options

melowe:


config-cli/src/main/java/com/quorum/tessera/config/cli/OverrideUtil.java
```diff
@@ -0,0 +1,116 @@
...
+import java.lang.reflect.Type;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Stream;
+import javax.xml.bind.annotation.XmlElement;
+
```
SatpalSandhu61: Can we have some documentation here please? I really shouldn't have to go through the code to try and work out what the class should do.
(TBH I can't really work out what it does)

namtruong (approved):

PR #343, opened 2018-08-01
Finalise CLI Config Migration utility

chris-j-h:
Testing and bugfixing for cli-config-migration utility


config-migration/src/main/java/com/quorum/tessera/config/builder/ConfigBuilder.java
```diff
@@ -322,11 +322,26 @@ public Config build() {
...
+        final List<Key> forwardingKeys;
+        if(alwaysSendTo != null) {
+            List<String> keyList = new ArrayList<>();
+
+            for(String keyPath : alwaysSendTo) {
+                try {
+                    String key = new String(Files.readAllBytes(Paths.get(keyPath)));
+                    keyList.add(key);
+                } catch (IOException e) {
+                    System.err.println("Error reading alwayssendto file: " + e.getMessage());
```
SatpalSandhu61, line 334: I'd suggest adding the filename & path in the error message.

config-migration/src/main/java/com/quorum/tessera/config/builder/KeyDataBuilder.java
```diff
@@ -42,15 +42,23 @@ public KeyDataBuilder withPrivateKeyPasswordFile(final Path privateKeyPasswordFi
...
     }
 
-    public KeyConfiguration build() {
-
+    public KeyConfiguration build(String workdir) {
         if(publicKeys.size() != privateKeys.size()) {
             throw new ConfigException(new RuntimeException("Different amount of public and private keys supplied"));
         }
 
+        if(publicKeys.size() == 0) return null;
```
SatpalSandhu61: It's good practise to put {} around the body of the 'if'.
Some code checkers (like sonar) will flag this up as a failure.

prd-fox (dismissed):

SatpalSandhu61 (dismissed):

melowe (approved):

config-migration/src/main/java/com/quorum/tessera/config/migration/TomlConfigFactory.java
```diff
@@ -83,7 +88,7 @@ public Config create(InputStream configData, InputStream... keyConfigData) {
         }
 
         KeyConfiguration keyData;
-        if(!publicKeyList.isEmpty() && !privateKeyList.isEmpty()) {
+        if(!publicKeyList.isEmpty() || !privateKeyList.isEmpty()) {
```
SatpalSandhu61: Is this change correct? Surely you only want to call the KeyDataBuilder.create() if both public and private keys exist? It will now call the create if only one of them is provided.

chris-j-h: The builder validates the lists to make sure they are the same size. If not it throws an exception with descriptive message. This seems preferable over silently overriding the users input with null.

SatpalSandhu61: Cool, that makes sense!

SatpalSandhu61 (approved):

PR #362, opened 2018-08-06
Just do key generation then exit if configfile not given

namtruong:
Make arg --configfile optional


prd-fox (dismissed):

Krish1979 (dismissed):

prd-fox (approved):

melowe (changes requested):
The logic in the else block assumes that the user can't define both configfile and keygen .

melowe (approved):

SatpalSandhu61 (approved):

PR #377, opened 2018-08-07
Logging changes

prd-fox:


namtruong (approved):

config/src/main/java/com/quorum/tessera/config/keys/KeyGeneratorImpl.java
```diff
@@ -39,17 +39,16 @@ public KeyGeneratorImpl(final NaclFacade nacl, final KeyEncryptor keyEncryptor,
     @Override
     public KeyData generate(final String filename) {
 
+        System.out.println("Enter a password if you want to lock the private key or leave blank");
```
SatpalSandhu61, line 42: add a comma after "lock the private key"?

PR #391, opened 2018-08-09
Read private keys correctly

prd-fox:


namtruong (dismissed):

config/src/main/java/com/quorum/tessera/config/adapters/KeyConfigurationAdapter.java
```diff
@@ -43,17 +45,26 @@ public KeyConfiguration unmarshal(final KeyConfiguration input) {
 
                     final KeyData kd = input.getKeyData().get(i);
 
+                    final KeyDataConfig keyData;
+
+                    if(kd.getConfig()==null) {
```
SatpalSandhu61, line 50: What if kd is null?

melowe (approved):

PR #397, opened 2018-08-12
Bugfix/dont validate in xmladapter

melowe:


config/src/main/java/com/quorum/tessera/config/adapters/KeyDataAdapter.java
```diff
@@ -37,10 +38,8 @@ public KeyData unmarshal(final KeyData keyData) {
             return unmarshalInline(keyData);
         }
 
-        if (keyData.getPublicKeyPath() == null || keyData.getPrivateKeyPath() == null
```
prd-fox, line 40: Need to keep the file exists checks here otherwise an IOException is thrown when trying to read the file, which results in JAXB not returning an element, and thus validation can't happen on this key after the configuration has been created.

prd-fox (changes requested):

config/src/main/java/com/quorum/tessera/config/adapters/KeyDataAdapter.java
melowe: I'll take a look at that case. Strongly suggest that the IO stuff needs to happen post unmarshalling.

prd-fox: Sure that's fine. Can do IO stuff after unmarshalling but pre-validation. But can leave for another day.

melowe: The keyData is returned and so any missing data should be flagged up by validations.

prd-fox (dismissed):

prd-fox (dismissed):

prd-fox (approved):

PR #400, opened 2018-08-13
Remove port from URL in configuration migration

prd-fox:
Check the port is a valid URL, and then only take the protocol and hostname from it, leaving the port out.


config-migration/src/main/java/com/quorum/tessera/config/migration/TomlConfigFactory.java
```diff
@@ -55,7 +42,16 @@ public Config create(InputStream configData, ArgonOptions options, String... fil
             LOGGER.debug("Found entry in toml file : {} {}", entry.getKey(), entry.getValue());
         });
 
-        final String url = toml.getString("url");
+        final String urlWithoutPost = Optional
```
namtruong: urlWithoutPort

config-migration/src/main/java/com/quorum/tessera/config/migration/TomlConfigFactory.java
```diff
@@ -144,7 +140,7 @@ public Config create(InputStream configData, ArgonOptions options, String... fil
 
         ConfigBuilder configBuilder = ConfigBuilder.create()
                 .serverPort(port)
-                .serverHostname(url)
+                .serverHostname(urlWithoutPost)
```
namtruong: same

namtruong (approved):

PR #405, opened 2018-08-13
Config migration bugfixes

chris-j-h:
* Fix null pointer exception if no tomlfile provided.  
* Add message for when password override is not applied when no keys are overriden
* If no workdir CLI override is provided then toml workdir is applied to all
* If workdir CLI override is provided then it is applied to all
  * Migration now only creates one instance of ConfigBuilder, initially setting with toml values and then cli overrides.


melowe (dismissed):

melowe (dismissed):

config-migration/src/main/java/com/quorum/tessera/config/builder/ConfigBuilder.java
```diff
@@ -153,22 +92,22 @@ public ConfigBuilder sslServerKeyStorePath(String sslServerKeyStorePath) {
...
-    public ConfigBuilder sslServerTrustStorePassword(String sslServerTrustStorePassword) {
-        this.sslServerTrustStorePassword = sslServerTrustStorePassword;
-        return this;
-    }
-
-    public ConfigBuilder sslServerKeyStorePassword(String sslServerKeyStorePassword) {
-        this.sslServerKeyStorePassword = sslServerKeyStorePassword;
-        return this;
-    }
+//    public ConfigBuilder sslServerTrustStorePassword(String sslServerTrustStorePassword) {
```
SatpalSandhu61: Should remove these if they're not needed.

SatpalSandhu61 (dismissed):

SatpalSandhu61 (dismissed):

prd-fox (dismissed):

SatpalSandhu61 (dismissed):

SatpalSandhu61 (dismissed):

namtruong (approved):

PR #427, opened 2018-08-17
Improvement: normalise shared code

melowe:
Move repeated functional interfaces and callback execute functions into shared location.


tests/jmeter-test/src/test/java/com/quorum/tessera/jmeter/PlaceHolder.java
```diff
@@ -0,0 +1,9 @@
+package com.quorum.tessera.jmeter;
```
SatpalSandhu61: Is this meant to be part of this PR?

config/217391d4-bdd1-4bd3-9364-1335d9522ba9.key
```diff
@@ -0,0 +1,6 @@
+{
```
SatpalSandhu61, line 1: Also not sure why these keys are being added. They don't appear to be used anywhere.

tests/jmeter-test/src/test/java/com/quorum/tessera/jmeter/PlaceHolder.java
melowe: Yes. That is the fix.

SatpalSandhu61: But this PR isn't for the jmeter fix - it's only meant to be a refactoring of the code into a shared module isn't it?

melowe: I see what you mean

PR #432, opened 2018-08-19
Improvement rationalise shared

melowe:
Move duplicated code, mainly functional interfaces and execution functions into shared location


config-cli/src/main/java/com/quorum/tessera/io/IOCallback.java
```diff
@@ -1,30 +0,0 @@
-
-package com.quorum.tessera.io;
-
-import java.io.IOException;
-import java.io.UncheckedIOException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-/**
- * Callback and template function t uncheck IO exceptions
```
SatpalSandhu61, line 9: typo

melowe: yes.. it appears to be a typo in a source file that no longer exists and this change enabled other changes I have backed up over three days.

prd-fox (approved):

PR #439, opened 2018-08-20
Add support for alwaysSendTo configuration. 

melowe:
1. Decouple NACL key from config object, use a String list to store keys
2. Use java bean validation 2.* as it allows to validate list elements 
	i) Use Base64 validator to validate always send to list. 
3. Change cli and cli migration modules to use string rather than key objects
4. Add KeysConvertor to convert List<String> to List<Key> when creating KeyManager instance


config/src/main/java/com/quorum/tessera/config/keys/KeysConverter.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Base64;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+
+public interface KeysConverter  {
+
+    static List<Key> convert(List<String> values) {
+        return  Objects.requireNonNull(values, "Key values cannot me null")
```
SatpalSandhu61: Typo

config/src/main/java/com/quorum/tessera/config/keys/KeysConverter.java
```diff
@@ -0,0 +1,20 @@
+package com.quorum.tessera.config.keys;
+
+import com.quorum.tessera.nacl.Key;
+import java.util.Base64;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+
```
SatpalSandhu61: Can we have a description of the class pls? What/Why is it converting keys...?

namtruong (approved):

PR #440, opened 2018-08-21
Feature additive override for peers

melowe:
Additional house keeping
1. Change key data validation to use validation api 2.0 rather than validate list
2. Add delegate setPosixFilePermissions to Files delegate 
3. Clean up tests that write to basedir (test jdbc url and key generator test)


config-cli/src/main/java/com/quorum/tessera/config/cli/OverrideUtil.java
```diff
@@ -28,10 +29,12 @@
 
 public interface OverrideUtil {
```
SatpalSandhu61, line 30: Can we have a header explaining what this class is for?
I shouldn't have to go through the code to work out the purpose of it.

Krish1979 (approved):

Issue #447, opened 2018-08-29
Private key password Encryption - Add ability to change password


Krish1979:
Strategically we will have key management tools integrated onto Tessera

prd-fox:
Requirements:
Private keys that are protected with a password should be able to have that password changed.

Outstanding questions:
- What command line parameters to use for specifying a password change?
- Should this operation update the configuration and then autostart the application, like the new key generation does?

prd-fox:
Answers to above questions:

a)
Use separate CLI options that specify:

1. The private key file to modify (inline changes will not be supported currently)
2. Provide the old password (if the key was locked) either directly or in a file
3. Provide argon options for that override the defaults

--key.pwdupdate.keyfile \<filepath\> [ --key.pwdupdate.old \<oldpass\> | --key.pwdupdate.pwdfile \<passwordFile\> ] [--key.pwdupdate.argon \<iterations\> \<memory\> \<parallelism\> ]

b)
This operation should overwrite the old key file as the only destructive operation. Updating the configuration with the new password (either the inline password list or the password file) is to be left to the end user.

melowe:
The update part can be a single flag and the same options can be used . 

```
--update-passwords or whatever the action is.. 
```

prd-fox:
Using:

--keys.passwordFile <file>
--keys.passwords <file>
--keys.keyData.privateKeyPath <path>
--keys.keyData.config.data.aopts.algorithm
--keys.keyData.config.data.aopts.iterations
--keys.keyData.config.data.aopts.memory
--keys.keyData.config.data.aopts.parallelism

Although they are longer and potentially less clear, they are already defined properties that can be interpreted in the context of updating passwords (instead of the normal case of overriding the configuration file).

prd-fox:
Implemented and the wiki updated.

PR #453, opened 2018-09-13
Move password reading

prd-fox:
Part of refactoring in preparation for https://github.com/jpmorganchase/tessera/issues/447

Move the user request for a password to the same class that the password is read from. 
This decouples the password reading from the key generation process.

### DIFF
--- a/config/src/main/java/com/quorum/tessera/config/util/PasswordReader.java
+++ b/config/src/main/java/com/quorum/tessera/config/util/PasswordReader.java
@@ -2,6 +2,7 @@ package com.quorum.tessera.config.util;
 
 import java.io.Console;
 import java.io.InputStream;
+import java.util.Objects;
 import java.util.Scanner;
 
 /**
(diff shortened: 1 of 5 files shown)


config/src/main/java/com/quorum/tessera/config/util/PasswordReader.java
```diff
@@ -23,7 +24,7 @@ public PasswordReader(final Console console, final InputStream systemIn) {
      * Read a password from either system console or the given stream
      * @return the read password, which may be empty if no password is given
      */
-    public String readPassword() {
+    public String readPasswordFromConsole() {
```
melowe, line 27: Can we make the PasswordReader an interface and then provide a CliPasswordReader or ConsolePasswordReader implementation? readPassword() is fine. So we can mock the password provider/reader and don't need to provide input streams when testing objects that require a PasswordReader

melowe (changes requested):

config/src/main/java/com/quorum/tessera/config/util/PasswordReader.java
prd-fox: Yes, that's a good idea.

The Console and CLI are the same thing in this case, it is just whether the JVM provides a Console object at runtime. So we would still need to fallback to the input stream as per the current implementation.

Just for tests we can use either a mocked (via Mockito) or test (via ServiceLoader) implementation.

Krish1979 (approved):

PR #456, opened 2018-09-14
Add private key support for Azure Key Vault

chris-j-h:
The private keys for a Tessera node can be stored in an Azure key vault instead of being provided in config or saved in files


chris-j-h:
Will close for the time being to add some additional features

PR #462, opened 2018-09-17
Add timestamp to encrypted transactions

chris-j-h:
New transactions are saved to encrypted transaction db table with timestamp in milliseconds since unix epoch


melowe (changes requested):
How do existing users upgrade to new db structure? We need the alter table sql scripts to accompany the release notes.

namtruong (approved):

PR #468, opened 2018-09-20
Improvement/centralise grpc logic

melowe:


prd-fox (approved):
Looks good to me. These changes won't affect Quorum integration until Quorum is ready for GRPC communication, and the acceptance tests pass.

Going to add an issue from this that the acceptance tests for GRPC should be expanded to have the same as the REST tests, all I can see right now is tests for /send and /version.

namtruong (approved):

PR #479, opened 2018-10-01
Feature/manual add peer

melowe:


prd-fox:
Adds groundwork for updatable configuration
Allow a new peer to be added to the configuration, where it can be picked up by services that use them (such as the WhitelistFilter).

prd-fox (approved):
Approved with the assumption that another change is required to enable the new peer to share its key information (PartyInfo).
As mentioned, this just lays some groundwork.

namtruong (approved):

PR #481, opened 2018-10-02
Add new key pair type for unsupported config combinations

chris-j-h:
- The UnsupportedKeyPair type has been added to account for keyData config combinations that do not fit into the pre-existing ConfigKeyPair implementations.
- Cleanup of KeyData related classes.  Much of this original behaviour is now encompassed in the individual ConfigKeyPair implementations.


config/src/main/java/com/quorum/tessera/config/keypairs/UnsupportedKeyPair.java
```diff
@@ -0,0 +1,84 @@
...
+        return privateKeyPath;
+    }
+
+    public KeyDataConfig getConfig() {
+        return config;
+    }
+
+    @Override
+    public void withPassword(String password) {
+
```
prd-fox: Add a comment here that this has intentionally been left blank as no implementation is needed

prd-fox (dismissed):

config/src/main/java/com/quorum/tessera/config/adapters/KeyDataAdapter.java
```diff
@@ -27,12 +26,44 @@ public ConfigKeyPair unmarshal(final KeyData keyData) {
...
+            FilesystemKeyPair kp = (FilesystemKeyPair) keyData;
+            return new KeyData(null, null, null, kp.getPrivateKeyPath(), kp.getPublicKeyPath());
+        }
+
+        if(type == UNSUPPORTED) {
+            UnsupportedKeyPair kp = (UnsupportedKeyPair) keyData;
+            return new KeyData(kp.getConfig(), kp.getPrivateKey(), kp.getPublicKey(), kp.getPrivateKeyPath(), kp.getPublicKeyPath());
+        }
+
+        throw new RuntimeException("The keypair type " + keyData.getClass() + " is not allowed");
```
melowe: Can we make this and illegal arguement, unsupported operation exception or some custom exception. We need to have a test for this scenario and we don't want to assert on a RuntimeException as null pointer will give you a positive outcome

melowe (approved):

prd-fox (approved):

PR #485, opened 2018-10-10
Improvement/enclave txnmgr

melowe:
1. Add clear separation between transaction management and encryption enabling the possibility to deploy enclave as a separate service co locating with private keys thus meaning tessera doesn't need to be co located nor have direct access to private keys. 

2. Fix issue where transaction manager/service we being used by enclave and not vice versa. 

3. Remove unnecessary transaction handling methods and objects. Make transaction manager methods strongly typed and not handling byte arrays. 

4. Add strongly typed keys rather than reusing a generic Key object. This makes the api and code read better but also potentially allows for the specialised handling for different key types. This also makes it easier to ensure that PrivateKeys aren't exposed beyond the enclave. 

5. Add specialised exceptions where RuntimeException was being bubbled up for better test cases.


tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoParser.java
```diff
@@ -90,14 +91,12 @@ default PartyInfo from(final byte[] encoded) {
         //so the prefix is always 2 (2 elements) and
         final List<byte[]> recipients = partyInfo.getRecipients()
             .stream()
-            .map(r -> {
```
prd-fox, line 93: This needs to be reversed for backwards compatibility.
The encode array method will add 8 bytes to the start of this field which shouldn't be there.

tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoServiceImpl.java
```diff
@@ -34,7 +34,7 @@ public PartyInfoServiceImpl(final PartyInfoStore partyInfoStore,
         this.partyInfoStore = Objects.requireNonNull(partyInfoStore);
         this.configService = Objects.requireNonNull(configService);
 
-        final String advertisedUrl = configService.getServerUri().toString() + "/";
```
prd-fox, line 37: Probably need a better way to do this, but we need to make sure the advertised URL ends with a "/" for backwards compatibility.
Maybe doing a check on the string to see if it ends with a slash, and if not, adding it? Instead of blindly adding the trailing slash (although the way we create the URI should mean it will always be needed).

tessera-core/src/main/java/com/quorum/tessera/sync/TransactionRequesterImpl.java
```diff
@@ -72,10 +74,11 @@ private boolean makeRequest(final String uri, final ResendRequest request) {
      * @param key the public key that transactions should be resent for
      * @return the request to be sent
      */
-    private ResendRequest createRequestAllEntity(final Key key) {
+    private ResendRequest createRequestAllEntity(final PublicKey key) {
 
         final ResendRequest request = new ResendRequest();
-        request.setPublicKey(key.toString());
+        String encoded = Base64.getEncoder().encodeToString(key.getKeyBytes());
+        request.setPublicKey(encoded);
```
prd-fox, line 81: Should use the Key method to encode it to string instead of getting an encoder on the bytes

prd-fox (changes requested):

tessera-core/src/main/java/com/quorum/tessera/sync/TransactionRequesterImpl.java
melowe: agreed.. but once we've stopped using the api objects as a domain model I hope that jaxb can deal with that sort of thing

tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoServiceImpl.java
melowe: I think this was lifted from existing code. But sure

tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoParser.java
melowe: not sure.. there should have been no changes to this. it sure why its showing up in this pr

PR #493, opened 2018-10-11
Merge/enclave txnmgr master

melowe:
1.Add clear separation between transaction management and encryption enabling the possibility to deploy enclave as a separate service co locating with private keys thus meaning tessera doesn't need to be co located nor have direct access to private keys.

2. Fix issue where transaction manager/service we being used by enclave and not vice versa.

3. Remove unnecessary transaction handling methods and objects. Make transaction manager methods strongly typed and not handling byte arrays.

4. Add strongly typed keys rather than reusing a generic Key object. This makes the api and code read better but also potentially allows for the specialised handling for different key types. This also makes it easier to ensure that PrivateKeys aren't exposed beyond the enclave.

5. Add specialised exceptions where RuntimeException was being bubbled up for better test cases.


prd-fox (approved):
Approved, but as discussed, the Enclave should not live inside the encryption module. This can be moved later, though.

PR #523, opened 2018-11-13
Improvement/communication config

melowe:


config-cli/src/test/resources/keygen-sample.json
```diff
@@ -5,10 +5,53 @@
...
+                "hostName": "http://localhost"
+            },
+            "communicationType" : "REST"
+        },
+        {
+            "app":"Q2T",
+            "enabled": true,
+            "serverSocket":{
+                "type":"UNIX",
+                "path":"/tmp/test.ipc"
```
SatpalSandhu61, line 24: Since the socket file is defined here now, is the definition for 'unixSocketFile' still needed (defined further below)?

SatpalSandhu61 (dismissed):

config-cli/src/test/resources/keygen-sample.json
prd-fox: Yes, at least for the time being, for backwards compatibility (users can still use the old configuration style).

config/src/main/java/com/quorum/tessera/config/KeyDataConfig.java
```diff
@@ -8,31 +8,30 @@
...
+    private PrivateKeyType type;
 
     public KeyDataConfig(PrivateKeyData privateKeyData, PrivateKeyType type) {
         this.privateKeyData = privateKeyData;
         this.type = type;
     }
 
-    private static KeyDataConfig create() {
-        return new KeyDataConfig(null, null);
+    public KeyDataConfig() {
```
prd-fox, line 30: Missing setters for fields - should add,  if only to be consistent with other configuration classes.

PR #536, opened 2018-11-22
Feature/merge raw tx support

nicolae-leonte-go:
merge the raw tx changes and a fix for the unix socket server app filtering

### DIFF
--- a/jaxrs-service/src/main/java/com/quorum/tessera/q2t/TransactionResource.java
+++ b/jaxrs-service/src/main/java/com/quorum/tessera/q2t/TransactionResource.java
@@ -70,6 +70,45 @@ public class TransactionResource {
 
     }
 
+    @ApiOperation(value = "Send private raw transaction payload", produces = "Encrypted payload hash")
+    @ApiResponses({
+        @ApiResponse(code = 200, response = SendResponse.class, message = "Send response"),
+        @ApiResponse(code = 400, message = "For unknown and unknown keys")
+    })
+    @POST
+    @Path("sendsignedtx")
+    @Consumes(APPLICATION_OCTET_STREAM)
+    @Produces(TEXT_PLAIN)
+    public Response sendSignedTransaction(
+        @HeaderParam("c11n-to") final String recipientKeys,
+        @NotNull @Size(min = 1) final byte[] signedTransaction) throws UnsupportedEncodingException {
+
+        SendSignedRequest sendSignedRequest = new SendSignedRequest();
+
+        sendSignedRequest.setHash(signedTransaction);
+
+        Optional.ofNullable(recipientKeys)
+            .filter(s -> !Objects.equals("", s))
+            .map(v -> v.split(","))
+            .ifPresent(sendSignedRequest::setTo);
+
+        final SendResponse response = delegate.sendSignedTransaction(sendSignedRequest);
+
+        final String encodedKey = response.getKey();
+
+        LOGGER.debug("Encoded key: {}", encodedKey);
+
+        java.net.URI location = UriBuilder.fromPath("transaction")
+            .path(URLEncoder.encode(encodedKey, StandardCharsets.UTF_8.toString()))
+            .build();
+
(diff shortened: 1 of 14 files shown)


jaxrs-service/src/main/java/com/quorum/tessera/q2t/TransactionResource.java
```diff
@@ -70,6 +70,45 @@ public Response send(
...
+            .map(v -> v.split(","))
+            .ifPresent(sendSignedRequest::setTo);
+
+        final SendResponse response = delegate.sendSignedTransaction(sendSignedRequest);
+
+        final String encodedKey = response.getKey();
+
+        LOGGER.debug("Encoded key: {}", encodedKey);
+
+        java.net.URI location = UriBuilder.fromPath("transaction")
```
prd-fox: Is the fully qualified name needed? It adds noise and should be removed if possible.

jaxrs-service/src/main/java/com/quorum/tessera/q2t/TransactionResource.java
```diff
@@ -70,6 +70,45 @@ public Response send(
...
+            .path(URLEncoder.encode(encodedKey, StandardCharsets.UTF_8.toString()))
+            .build();
+
+        //TODO: Quorum expects only 200 responses. When Quorum can handle a 201, change to CREATED
```
prd-fox, line 106: Is this called from Quorum? I haven't seen any changes for that and believe this is to be called from a 3rd party application. If so, this should return a 201 with the TODO resolved.

jaxrs-service/src/main/java/com/quorum/tessera/thridparty/RawTransactionResource.java
```diff
@@ -41,7 +41,7 @@ public RawTransactionResource(RawTransactionManager delegate) {
         @ApiResponse(code = 400, message = "For unknown sender")
     })
     @POST
-    @Path("store")
+    @Path("storeraw")
```
prd-fox, line 44: What was the reasoning for changing this? Not that it's wrong this way, just interested on the discussion points.

tessera-core/src/main/java/com/quorum/tessera/api/model/SendSignedRequest.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Used when a new transaction is to be created where this node is the sender
+ */
+@ApiModel
+public class SendSignedRequest {
+
+    @XmlMimeType("base64Binary")
+    @Size(min = 1)
+    @NotNull
+    @ApiModelProperty("Encrypted payload to send to other parties.")
```
prd-fox: Is this a hash or a payload? Assuming its a hash of the payload, in which case the @ApiModelProperty string should be updated.

prd-fox: Additionally, assuming it is a hash, then the size is fixed and the annotation can be updated accordingly. I believe it should always be 64 bytes, but you should double check.

jaxrs-service/src/main/java/com/quorum/tessera/q2t/TransactionResource.java
nicolae-leonte-go: imported java.net.URI

tessera-core/src/main/java/com/quorum/tessera/api/model/SendSignedRequest.java
nicolae-leonte-go: Updated the description. Since there is no restriction on ReceiveRequest.key (just min size 1) I prefer not do add such a constraint only for SendSignedRequest.

prd-fox: Fair enough. I think the same applies there but we should update these together, which can be done later.

prd-fox (approved):

PR #551, opened 2018-11-29
Add auto create table config option

melowe:
Added auto create tables config option to allow backward compatibility with existing config. 
Default create table in jpa config to none with aforementioned override. 
Package ddls into jar for improved visibility and potential use in automated builds.


melowe:
Dependent quorum-examples PR https://github.com/jpmorganchase/quorum-examples/pull/142

ddls/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+            </plugin>
+            
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+              
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <mainClass>dbsetup.Main</mainClass>
```
prd-fox: I can't see that this class exists. Should this plugin be here?

prd-fox (approved):

Issue #552, opened 2018-11-30
Upcheck works even when database tables aren't setup correctly


melowe:
Upcheck and other rest calls return ok status regardless of database state. For an upcheck to be successful some validation of the DB state should happen and an appropriate error status returned so fail early rather than when attempting to access data. 

Action: Add validation query to relevant status check endpoints e.g. select count(*) from table name or something similar.

Krish1979:
include the public keys of the Tessera node on the response too..

PR #564, opened 2018-12-05
Change `"I'm up!"` to `"TM_READY"` for easy discovery

vietlq:
Change `"I'm up!"` to `"TM_READY"` for easy discovery & safe search. Return codes should use limited char set (alpha numeric, upper case and underscore). Having `'`, mixed casing and punctuation complicates return codes.

Related to https://github.com/jpmorganchase/quorum-examples/issues/146


Krish1979:
as already confirmed in jpmorganchase/quorum-examples#146 this change is  not required

Krish1979:
also we are looking to implement #552 which will improve upcheck coverage scope.

PR #565, opened 2018-12-05
Add support for Hashicorp Vault to store Tessera keys

chris-j-h:
Add Hashicorp Vault to supported key vaults that can be used with Tessera.  Features include:

- Store generated public/private key pairs in a Vault configured with a version 2 k/v secret engine
- Retrieve public/private key pairs and previous versions of key pair from a pre-configured Vault
- TLS communication between Tessera and the Vault server is supported.  Non-TLS communication is also allowed for development/test purposes
- Vault config can be provided in configfile, and the corresponding CLI options have been added to provide this information for key generation
- `GetSecretData` and `SetSecretData` added as arguments for the `KeyVaultService` methods. This allows for the same method signatures to be used in the Hashicorp and Azure implementations despite each provider's different interpretation of what constitutes a secret


prd-fox (dismissed):

melowe (dismissed):

namtruong (dismissed):

chris-j-h:
Changes have been made to support v2 of the k/v secrets engine (i.e. versioning of secrets and retrieval of previous versions of a secret)

melowe (approved):

prd-fox (approved):

Issue #583, opened 2018-12-31
Update /config/peers as part of node to node interaction


melowe:
`/config/peers ` we only added as a means for verifying add peer invocations made by the admin cli. While this verification should be distinct as an admin service, however providing a read only path to list all the peers in a current network (including those added during party info dialogues) is desirable.

prd-fox:
Adding some extra parameters that display the "last connected" time, and number of failed requests to the target. The aim is to give the user some extra context to determine if the peer is offline or not.

PR #584, opened 2018-12-31
Fetch network information ("PartyInfo")

prd-fox:
Add an endpoint to fetch the contents of the "PartyInfo" store without having to supply your own (as in the usual peer exchange).

Fixes https://github.com/jpmorganchase/tessera/issues/583


prd-fox:
Going to change the structure on the output message.

SatpalSandhu61 (approved):

jaxrs-service/src/main/java/com/quorum/tessera/p2p/PartyInfoResource.java
```diff
@@ -56,8 +54,45 @@ public Response partyInfo(@ApiParam(required = true) final byte[] payload) {
...
-            .entity(streamingOutput)
-            .build();
+        return Response.status(Response.Status.OK).entity(streamingOutput).build();
     }
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Fetch network/peer information", produces = "public list of peers/publickey mappings")
+    @ApiResponses({@ApiResponse(code = 200, message = "Peer/Network information", response = PartyInfo.class)})
+    public Response getPartyInfo() {
```
nicolae-leonte-go, line 64: Is this a good time to consider adding a "monitoring" app (where the new service is exposed)? Although it is easy to test in most scenarios (where nobody bothers with P2P app encryption) when the P2P app is secured it may prove harder to invoke with tools like curl (where a java key store may not be immediately usable).

nicolae-leonte-go (approved):

jaxrs-service/src/main/java/com/quorum/tessera/p2p/PartyInfoResource.java
prd-fox: Yep, a monitoring app would certainly be useful. InfluxDB currently captures a few metrics, but this is only for the Jersey/Grizzly HTTP server, and so should be expanded (and other info should be included aside from metrics).

PR #585, opened 2019-01-02
Add a "lastContacted" field to the peers list

prd-fox:
Add a `lastContacted` field to the peers list when fetching the party info from `/partyInfo` , which gives the timestamp of when the peer was last called to/from.
It is intended to give an indication of whether the peer is currently active or not. The older the timestamp, the less likely it is that peer is online/reachable.
If the `lastContacted` field is `null` then it means no communication has ever occurred with that peer.

A (truncated) sample output:
```
{
    "url": "http://localhost:9001/",
    "peers": [
        {
            "url": "http://localhost:9006/",
            "lastContact": "2019-01-02T13:49:01.095Z"
        },
        {
            "url": "http://localhost:9005/",
            "lastContact": "2019-01-02T13:49:01.098Z"
        },
        {
            "url": "http://localhost:9007/",
            "lastContact": null
        }
    ],
    "keys": [
        {
            "key": "R56gy4dn24YOjwyesTczYa8m5xhP6hF2uTMCju/1xkY=",
            "url": "http://localhost:9005/"
        },
        {
            "key": "ROAZBWtSacxXQrOe3FGAqJDyJjFePR5ce4TSIzmJ0Bc=",
            "url": "http://localhost:9007/"
        },
        {
            "key": "UfNSeSGySeKg11DVNEnqrUtxYRVor4+CvluI8tVv62Y=",
            "url": "http://localhost:9006/"
        }
    ]
}
```


tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoStore.java
```diff
@@ -39,6 +40,12 @@ public PartyInfoStore(final ConfigService configService) {
     public synchronized void store(final PartyInfo newInfo) {
         recipients.addAll(newInfo.getRecipients());
         parties.addAll(newInfo.getParties());
+
+        //update the sender to have been seen recently
+        final Party sender = new Party(newInfo.getUrl());
```
SatpalSandhu61, line 45: There's a small issue here. If the url has been added to the config without a trailing '/' (like in 7node examples) then we end up with duplicate entries, e.g.:
'''
<snip>
"peers":[
        {"url":"http://localhost:9007","lastContact":null},
        {"url":"http://localhost:9005","lastContact":null},
        {"url":"http://localhost:9006","lastContact":null},
        {"url":"http://localhost:9003","lastContact":null},
        {"url":"http://localhost:9004","lastContact":null},
        {"url":"http://localhost:9001","lastContact":null},
        {"url":"http://localhost:9002","lastContact":null},
        {"url":"http://localhost:9006/","lastContact":"2019-01-03T03:13:38.984Z"},
        {"url":"http://localhost:9005/","lastContact":"2019-01-03T03:13:38.984Z"},
        {"url":"http://localhost:9004/","lastContact":"2019-01-03T03:13:38.985Z"},
        {"url":"http://localhost:9003/","lastContact":"2019-01-03T03:13:38.986Z"},
        {"url":"http://localhost:9002/","lastContact":"2019-01-03T03:13:38.986Z"},
        {"url":"http://localhost:9001/","lastContact":"2019-01-03T03:13:38.982Z"},
        {"url":"http://localhost:9007/","lastContact":"2019-01-03T03:13:38.987Z"}
],
<snip>
'''
I think the root cause of this might be in the constructor for PartyInfoServiceImpl, where the initialParties value is taken from the configured peers.
I also notice that we have a bit of duplicate code dotted around to add the trailing forward-slash. Maybe it could all be consolidated into PartyInfoStore?

SatpalSandhu61: BTW after posting my comment, I pulled the latest 7nodes example and noticed that has been changed to include the forward-slash on the config.
However, we should still modify the Tessera code so that with/without forward-slash are treated the same.

prd-fox: I will filter out the URLs without a /, as these will always be null.

SatpalSandhu61 (approved):

PR #594, opened 2019-01-11
Initial groundwork to deploy enclave as distinct service.

melowe:
Create remotable Enclave implementation EnclaveClient .

Changes are dormant for the time being as there are a lot of modules impacted. 
Moved payload encoders into encryption module. 

Add enclave as server/app type to config. 

Throw illegal state exception when cli delegate hasn't been initialised 
but attempts to access config is made.


prd-fox:
LGTM.

Talked about moving enclave related pieces to own module, away from the nacl encryption code, otherwise its okay.

prd-fox (approved):

Issue #579, opened 2018-12-27
Resend should also restore missing transactions for an originating node


SatpalSandhu61:
Currently, if a node loses its database then transactions are restored by means of the other nodes performing a resend for transactions where it is a participant.
However, if the failed node is the originator of the transaction, then those transactions won't be restored. See issue # [607](https://github.com/jpmorganchase/quorum/issues/607).
This represents a flaw in the restore mechanism and means that users would have to build their own mechanism to store transactions and restore them after a node failure.

We should therefore extend the resend so that transactions are also restored to the originating node.

Note that Tessera stores the sender of a transaction, so that could be used in order to restore transactions back to the originating node.

prd-fox:
Using last comment above as a starting point, the solution would be:

1. Resend the tx from a receiving node back to the originator (using a new endpoint to identify the nature of this resend, maybe?)
2. Recreate the tx in the originating node as if it has no participants
3. Add the recipient that sent us this as one of the recipients
4. If another node sends us the same tx, add it as a recipient too, so we can rebuild the list of participants.

Note that if note that if both the originator AND a recipient both lose the data, then it becomes a lot harder to recover this, and will not be included in the scope of this issue. It would require some verification that the recipient was indeed an original party (probably using some data on the blockchain).

PR #608, opened 2019-01-21
Rebuild own transaction list

prd-fox:
When we lose transactions fro the database, e.g. corruption or external rollback, then we need to be able to reconstruct any transactions that may be missing.

In the case where we created the transaction, we must rebuild this list based on the participants telling us as such.

Changes:
- when searching for transactions for a resend request, keep messages where the sender was the requested search key (in addition to the existing condition that is it a recipient)
- when receiving the message to store (a `push` request), then check if the message has our own key as the sender, and if so, begin the reconstruction process.

The reconstruction process is as follows:
- If the tx doesn't exist, add our own key and recipientBox to the payload, then store it in the database.
- If the tx does exist, and doesn't contain the recipient who sent it to us, store* the public key and recipientBox on the payload and update.
- If the tx does exist, and the recipient is already present, do nothing.

*if we receive a malicious message, i.e. adding a sender to message who wasn't originally there, then this is handled. We add the recipient and box still, but since we did not recalculate the box, only store what they gave us, then we would only store a box that doesn't work. When we resend the message back to the malicious party, then we would only resend the non-working box, so it won't be able to be decrypted.

Fixes https://github.com/jpmorganchase/tessera/issues/579


tessera-core/src/main/java/com/quorum/tessera/transaction/ResendManagerImpl.java
```diff
@@ -0,0 +1,85 @@
...
+        final Optional<EncryptedTransaction> tx = this.encryptedTransactionDAO.retrieveByHash(transactionHash);
+
+        if (tx.isPresent()) {
+
+            //we just need to add the recipient
+            final byte[] encodedPayload = tx.get().getEncodedPayload();
+            final EncodedPayload existing = payloadEncoder.decode(encodedPayload);
+
+            if (!existing.getRecipientKeys().contains(payload.getRecipientKeys().get(0))) {
+                //You could check that the provided information really does decrypt the payload
```
nicolae-leonte-go: I think this check should actually be implemented. It is not hard to see a "malicious" party which sends bogus garbage for every unknown transaction (not sure how easy it is to produce a byte array which when hashed results in a certain hash). In such a case that party would be added to the recipients list which would then allow it to invoke the resend and then gain access to the actual encoded payload - which would then facilitate a brute force attack.
The check should be that decrypting the box produces the expected master key, and that the decrypted transaction payload from the input transaction matches the decrypted  payload from the DB.

tessera-core/src/main/java/com/quorum/tessera/transaction/TransactionManagerImpl.java
```diff
@@ -153,41 +161,54 @@ public SendResponse sendSignedTransaction(SendSignedRequest sendRequest) {
...
+                .map(EncryptedTransaction::getEncodedPayload)
+                .map(payloadEncoder::decode)
+                .filter(payload -> {
+                    final boolean isRecipient = payload.getRecipientKeys().contains(recipientPublicKey);
+                    final boolean isSender = Objects.equals(payload.getSenderKey(), recipientPublicKey);
+                    return isRecipient || isSender;
+                }).forEach(payload -> {
+                    if (Objects.equals(payload.getSenderKey(), recipientPublicKey)) {
+                        final PublicKey decryptedKey = searchForRecipientKey(payload).orElseThrow(RuntimeException::new);
+                        payload.getRecipientKeys().add(decryptedKey);
```
melowe, line 191: Can we have a descriptive exception type. Looks like a NoKeyFound or something.

tessera-core/src/main/java/com/quorum/tessera/transaction/ResendManagerImpl.java
prd-fox: I'm happy to add the check. I think a brute force attack would still be infeasible, but better to make it entirely impossible.

prd-fox: Added a guard that checks the cipher text and decrypted payloads match.

melowe (dismissed):

SatpalSandhu61 (dismissed):

nicolae-leonte-go (dismissed):

melowe (dismissed):

melowe (approved):

PR #616, opened 2019-01-25
Automate deploy and releases

melowe:
Add build hooks and initial script to position for 
1. Deploy snapshot when merge to master
2. Create release when build is invoked via api with the job name "release"

This change prints variables to ensure we have enough to infer if a deploy or release. No deploy or release is invoked.


.travis.yml
```diff
@@ -11,24 +11,34 @@ jdk:
...
 - mkdir -p vault/bin && pushd $_
 - unzip /tmp/vault_1.0.1_linux_amd64.zip
 - export PATH=$PATH:$PWD && popd
 
-script:
-- mvn install -P acceptance-test,vault-acceptance-tests,jnacl
-
-before_cache:
-- mvn dependency:purge-local-repository -Dinclude=com.quorum.tessera -DresolutionFuzziness=groupId -Dverbose=true -DreResolve=false
+script: mvn install -P acceptance-test,vault-acceptance-tests,jnacl -DskipTests=true
```
prd-fox: We shouldn't skip tests, even if it is to try and speed up the build for the release purpose

prd-fox (changes requested):

.travis.yml
melowe: mvn test gets executed anyway.. although I'll check this is happening. I do need to check that the acceptance tests are run.

PR #623, opened 2019-02-06
Enclave over WebSockets

melowe:
Currently dormant module enabling the remoting of enclave over WebSockets in the same way as enclave-jaxws remotes over rest/http. EnclaveEndpointTest fires up local WebSockets server and delegates to mocked enclave instance.


enclave/enclave-api/src/main/java/com/quorum/tessera/enclave/AlwaysSendToHelper.java
```diff
@@ -0,0 +1,27 @@
+package com.quorum.tessera.enclave;
```
prd-fox: What issue is this really addressing? What was wrong about the configuration passing in the list directly? 

As far as I can see, the configuration object is still in scope for both the main application and the standalone enclave, so it's not a scoping issue. This also just sits in-between the config and the key converter, so it doesn't seem to be a circular reference issue either.

enclave/enclave-api/src/main/resources/tessera-enclave-spring.xml
```diff
@@ -17,7 +17,19 @@
...
 
+    <bean id="environmentVariableProvider" class="com.quorum.tessera.config.util.EnvironmentVariableProvider"/>
 
+    <bean name="keyPairConverter" class="com.quorum.tessera.keypairconverter.KeyPairConverter">
+        <constructor-arg ref="config"/>
+        <constructor-arg ref="environmentVariableProvider"/>
+    </bean>
+    
+
+    <bean id="alwaysSendToHelper" class="com.quorum.tessera.enclave.AlwaysSendToHelper">
```
prd-fox: Pending above issue, this could be removed.

enclave/enclave-api/src/test/java/com/quorum/tessera/enclave/AlwaysSendToHelperTest.java
```diff
@@ -0,0 +1,20 @@
+package com.quorum.tessera.enclave;
```
prd-fox: Pending above issue, this could be removed.

enclave/enclave-websockets/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+            <type>jar</type>
+        </dependency>
+        
+        <dependency>
+            <groupId>com.jpmorgan.quorum</groupId>
+            <artifactId>enclave-api</artifactId>
+        </dependency>
+        
+                
+        <dependency>
```
prd-fox, line 32: Is there a reason use this server instead of the Jersey-Grizzly test server used elsewhere?

enclave/enclave-websockets/src/main/java/com/jpmorgan/quorum/enclave/websockets/EnclaveAdapter.java
```diff
@@ -0,0 +1,182 @@
...
+    public byte[] createNewRecipientBox(EncodedPayload payload, PublicKey recipientKey) {
+        webSocketTemplate.execute(s -> {
+            EnclaveRequest request = EnclaveRequest.Builder.create()
+                    .withType(EnclaveRequestType.CREATE_NEW_RECIPIENT_BOX)
+                    .withArg(payload)
+                    .withArg(recipientKey).build();  
+            s.getBasicRemote().sendObject(request);
+        });
+        
+        return client.pollForResult(ByteBuffer.class).get().array();
```
prd-fox, line 179: Are you sure that a ByteBuffer is the returned object? It looks like under the hood it sends the byte array, which is what we should be receiving.

prd-fox: Better yet, should we encode it before sending over the wire? e.g. Base64 or something, then we can simply send a string an not worry about this issue.

enclave/enclave-websockets/src/main/java/com/jpmorgan/quorum/enclave/websockets/EnclaveEndpoint.java
```diff
@@ -0,0 +1,149 @@
...
+    @OnMessage
+    public void onRequest(Session session, EnclaveRequest request) {
+
+        LOGGER.info("Request {}", request.getClass());
+
+        EnclaveRequestType type = request.getType();
+
+        final WebSocketTemplate webSocketTemplate = new WebSocketTemplate(session);
+
+        if (type == EnclaveRequestType.DEFAULT_PUBLIC_KEY) {
```
prd-fox: Are `if` statements really the best option here? A `switch` statement I think it the better alternative, either directly or passing off to private methods.
This method is doing too many things otherwise.

enclave/enclave-websockets/src/main/java/com/jpmorgan/quorum/enclave/websockets/EnclaveEndpoint.java
```diff
@@ -0,0 +1,149 @@
...
+            EncodedPayload payload = (EncodedPayload) request.getArgs().get(0);
+            PublicKey recipientKey = (PublicKey) request.getArgs().get(1);
+            
+            byte[] boxData = enclave.createNewRecipientBox(payload, recipientKey);
+            webSocketTemplate.execute((s) -> {
+                s.getBasicRemote().sendBinary(ByteBuffer.wrap(boxData));
+            });
+            return;
+        }
+        throw new UnsupportedOperationException(type + " is not supported");
```
prd-fox: Can we send back an error to the caller?

enclave/enclave-websockets/src/main/java/com/jpmorgan/quorum/enclave/websockets/EnclaveRequestCodec.java
```diff
@@ -0,0 +1,136 @@
...
+        EnclaveRequestType enclaveRequestType = request.getType();
+
+        JsonArrayBuilder jsonArrayBuilder = Json.createArrayBuilder();
+
+        for (int i = 0; i < enclaveRequestType.getParamTypes().size(); i++) {
+            Object value = request.getArgs().get(i);
+
+            ArgType type = enclaveRequestType.getParamTypes().get(i);
+
+            if (type == ArgType.PUBLIC_KEY_LIST) {
```
prd-fox: A `switch` would be better here.

prd-fox (changes requested):

Krish1979 (approved):

prd-fox (approved):

PR #635, opened 2019-02-20
Retrieve node key pair information using admin API

chris-j-h:
* Add `adminhost:port/config/keypairs` endpoint to retrieve all key pairs for a particular node.  Example response is:
```json
[
   {
      "publicKey" : "oNspPPgszVUFw0qmGFfWwh1uxVUXgvBxleXORHj07g8="
   },
   {
      "publicKey" : "ABn6zhBth2qpdrJXp98IvjExV212ALl3j4U//nj4FAI="
   }
]
```
*  Add `adminhost:port/config/keypairs/{publickey}` endpoint to check if node is registered with a particular key pair.  Example response is:
```json
  {
      "publicKey" : "oNspPPgszVUFw0qmGFfWwh1uxVUXgvBxleXORHj07g8="
   }
```
if the node is registered with the key pair, and:
```
No key pair found with public key oNspPPgszVUFw0qmGFfWwh1uxVUXgvBxleXORHj07g8=
```
if node is not registered with the key pair

### DIFF
--- a/jaxrs-service/src/main/java/com/quorum/tessera/admin/ConfigResource.java
+++ b/jaxrs-service/src/main/java/com/quorum/tessera/admin/ConfigResource.java
@@ -1,6 +1,5 @@
 package com.quorum.tessera.admin;
 
-import com.quorum.tessera.config.KeyData;
 import com.quorum.tessera.config.Peer;
 import com.quorum.tessera.config.apps.AdminApp;
 import com.quorum.tessera.core.config.ConfigService;
@@ -104,10 +103,8 @@ public class ConfigResource implements AdminApp {
             throw new NotFoundException("No key pair found with public key " + base64PublicKey);
         }
 
-        //TODO use dedicated transport object instead?
-        KeyData responseData = new KeyData();
+        PublicKeyResponse responseData = new PublicKeyResponse();
         responseData.setPublicKey(base64PublicKey);
-        responseData.setPrivateKey("REDACTED");
 
         return Response.ok(responseData).build();
     }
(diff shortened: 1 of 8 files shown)


jaxrs-service/src/main/java/com/quorum/tessera/admin/ConfigResource.java
```diff
@@ -85,4 +90,49 @@ public Response getPeers() {
...
+
+        return Response.ok(responseData).build();
+    }
+
+    @GET
+    @Path("/keypairs")
+    public Response getKeyPairs() {
+        Set<PublicKey> publicKeys = configService.getPublicKeys();
+
+        if(publicKeys.isEmpty()) {
```
prd-fox: It would be better for return the empty list instead of throwing an error here; the resource exists, but is just empty.

jaxrs-service/src/main/java/com/quorum/tessera/admin/ConfigResource.java
```diff
@@ -85,4 +90,49 @@ public Response getPeers() {
...
+            throw new NotFoundException("No key pairs found");
+        }
+
+        //TODO use dedicated transport object instead?
+        List<KeyData> keyPairData = publicKeys.stream()
+            .map(PublicKey::encodeToBase64)
+            .map(publicKey -> {
+                KeyData kd = new KeyData();
+                kd.setPublicKey(publicKey);
+                kd.setPrivateKey("REDACTED");
```
prd-fox: How about creating a new object that only contains the public key? The "REDACTED" and type in the response isn't very useful for the user.

prd-fox (changes requested):

jaxrs-service/src/main/java/com/quorum/tessera/admin/ConfigResource.java
chris-j-h: This was something I had considered doing - I'll add it in

prd-fox: Noticed you mentioned the same for the single key endpoint. I think it is worth doing; if you don't fancy using a new object you can create the JSON response directly.

prd-fox (changes requested):

jaxrs-service/src/main/java/com/quorum/tessera/admin/PublicKeyResponse.java
```diff
@@ -0,0 +1,20 @@
...
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+class PublicKeyResponse {
+
+    @XmlElement
+    private String publicKey;
```
prd-fox, line 11: This element should have a getter too, even though JAX-RS doesn't need it.

prd-fox (approved):

PR #636, opened 2019-02-20
Standalone enclave JAXRS and Websockets

melowe:
- Added server container to enclave to monitor status and stop start service as required. When there is a connection issue the service container will check at a given interval . 
- Transaction manager checks enclave status before starting to execute business functions thus narrowing the window for errors related to connecting to enclave instances. 
- Enclave instances are now more strongly encapsulated behind a Factory as a move toward modularisation. When tessera sees there is an enclave server configured it creates the relevant client to connect to a standalone enclave.


enclave/enclave-api/src/main/java/com/quorum/tessera/enclave/EnclaveClientFactory.java
```diff
@@ -0,0 +1,16 @@
...
+import com.quorum.tessera.config.Config;
+import java.util.ServiceLoader;
+
+
+public interface EnclaveClientFactory<T extends EnclaveClient> {
+
+    T create(Config config);
+    
+    static EnclaveClientFactory create() {
+        return ServiceLoader.load(EnclaveClientFactory.class)
```
prd-fox, line 12: Could this use the `ServiceLoaderUtil`?

enclave/enclave-api/src/main/java/com/quorum/tessera/enclave/EnclaveFactory.java
```diff
@@ -1,13 +1,45 @@
...
+import com.quorum.tessera.keypairconverter.KeyPairConverter;
+import com.quorum.tessera.nacl.NaclFacadeFactory;
+import java.util.Collection;
+import java.util.Optional;
 
-public interface EnclaveFactory<C> {
+public interface EnclaveFactory {
 
-    Enclave create(C config);
+    static Enclave createServer(Config config) {
```
prd-fox, line 18: This seems be the default enclave, rather than the server. I think this should be named `createDefaultEnclave` or something similar.

enclave/enclave-api/src/test/java/com/quorum/tessera/enclave/EnclaveFactoryTest.java
```diff
@@ -18,5 +29,70 @@ public void onSetUp() {
     public void create() {
         assertThat(enclaveFactory).isNotNull();
     }
+ @Test
```
prd-fox: Whitespacing

enclave/enclave-jaxrs/src/main/java/com/quorum/tessera/enclave/rest/RestfulEnclaveClient.java
```diff
@@ -173,10 +172,27 @@ public RawTransaction encryptRawPayload(byte[] message, PublicKey sender) {
...
     }
+
+    @Override
+    public Status status() {
+        
+        final Response response = client.target(uri)
+                .path("ping")
+                .request().get();
+        
+        if(response.getStatus() == 200) {
```
prd-fox, line 189: Not sure this is right. The `/ping` endpoint just returns the state of the remote Enclave. So if the Enclave has `STOPPED` status, the request would still be a `200`. Seems like the response needs to be evaluated back to the enum.

prd-fox (dismissed):

enclave/enclave-jaxrs/src/main/java/com/quorum/tessera/enclave/rest/RestfulEnclaveClient.java
melowe: We only use the http code to communicate state.

enclave/enclave-api/src/main/java/com/quorum/tessera/enclave/EnclaveFactory.java
melowe: Default has the same issue. I think we should rename to something else, but at least server has symmetry with 'client' for the cases where we're remoting. Either wa can be deferred as  a minor change in another change set.

prd-fox (approved):

PR #647, opened 2019-02-28
Allow for TLS keystore passwords to be given as environment variables

chris-j-h:
When currently configuring P2P TLS, the passwords for client and server keystore/truststores can only be provided as part of the json configfile.  Ideally, credentials should not be stored in files and so this PR allows for the passwords to be provided by setting environment variables.

#### Precedence
1. If the _prefixed_ environment variable is set this will be used, else
1. If the config value is set this will be used, else
1. If the _global_ environment variable is set this will be used

#### Global TLS environment variables
If set, these are applied to all server configs defined in the configfile.  i.e. if a P2P and ADMIN server are both configured with TLS then the values set for the global environment variables will be used for both.  The global environment variables are:
* `TESSERA_SERVER_KEYSTORE_PWD`
* `TESSERA_SERVER_TRUSTSTORE_PWD`
* `TESSERA_CLIENT_KEYSTORE_PWD`
* `TESSERA_CLIENT_TRUSTSTORE_PWD`

These values are ignored if the passwords are also provided in the configfile or prefixed environment variables. 

#### Prefixed TLS environment variables
An optional field, `environmentVariablePrefix`, has been added to `sslConfig`, e.g.:
```
"sslConfig": {
        ...
        "serverKeyStore": ...,
        "serverTrustStore": ...,
        "clientKeyStore": ...,
        "clientTrustStore": ...,
        "environmentVariablePrefix": "PREFIX"
         ...
},
```
The passwords for the keystores specified in the `sslConfig` can be set using the prefixed environment variables:
* `PREFIX_TESSERA_SERVER_KEYSTORE_PWD`
* `PREFIX_TESSERA_SERVER_TRUSTSTORE_PWD`
* `PREFIX_TESSERA_CLIENT_KEYSTORE_PWD`
* `PREFIX_TESSERA_CLIENT_TRUSTSTORE_PWD`

Using unique `environmentVariablePrefix` values for each TLS server configured allows for the keystore passwords for each server to be set.


config/src/main/java/com/quorum/tessera/config/util/EnvironmentVariableProvider.java
```diff
@@ -15,4 +15,10 @@ public String getEnv(String name) {
             .orElse(null);
     }
 
+    public boolean hasEnv(String name) {
+        Optional<String> env = Optional.ofNullable(System.getenv(name));
```
prd-fox: Seems a bit much for a null check. Is `return Objects.isNull(getEnv(name));` enough?

prd-fox (changes requested):

config/src/main/java/com/quorum/tessera/config/util/EnvironmentVariables.java
```diff
@@ -0,0 +1,14 @@
+package com.quorum.tessera.config.util;
+
+public final class EnvironmentVariables {
+
+    public static final String serverKeyStorePwd = "TESSERA_SERVER_KEYSTORE_PWD";
```
melowe: java naming convention is static final in uppercase with underscore word separators.

config/src/main/java/com/quorum/tessera/config/util/EnvironmentVariableProvider.java
chris-j-h: I have kept this as-is for the time being as it is a workaround for the jacoco coverage report.  `System.getenv()` cannot be mocked and it is not trivial to set the environment variables when testing.  If we were to use `Objects.isNull(...)`, then both the `null` and non-`null` cases would need to be tested to satisfy jacoco.  Using the Optional means we achieve coverage by only testing the `null` case.

melowe (approved):

melowe (approved):

melowe (dismissed):

prd-fox (approved):

PR #671, opened 2019-03-27
Align data migration tool with current version

melowe:
Discovered issue where constellation data store migration tool is missing TIMESTAMP column. 
Added `TIMESTAMP` column definition when performing the data migration, which means the created database is compliant with the latest release version.

Use provided DDL scripts for acceptance tests (H2, HSQL, and sqlite) to ensure these are aligned with current code base. 

Added HSQL ddl.


data-migration/src/main/java/com/quorum/tessera/data/migration/H2DataExporter.java
```diff
@@ -1,21 +1,20 @@
...
-import java.util.Map.Entry;
 
 public class H2DataExporter implements DataExporter {
 
     private static final String INSERT_ROW = "INSERT INTO ENCRYPTED_TRANSACTION (HASH,ENCODED_PAYLOAD) VALUES (?,?)";
 
-    private static final String CREATE_TABLE = "CREATE TABLE ENCRYPTED_TRANSACTION "
-        + "(ENCODED_PAYLOAD LONGVARBINARY NOT NULL, "
-        + "HASH LONGVARBINARY NOT NULL UNIQUE, PRIMARY KEY (HASH))";
+    private static final String CREATE_TABLE = "CREATE TABLE " +
```
prd-fox: Like you've done for the tests, could we not use the DDL scripts to do the creation? It would prevent future issues - maybe even create some `INSERT` DDL scripts, although this would be specific for the migration, so may not be appropriate.

prd-fox (dismissed):

data-migration/src/main/java/com/quorum/tessera/data/migration/H2DataExporter.java
prd-fox: Also applies to other DB types, but won't duplicate the comment.

prd-fox: Also, didn't we provide the ability for custom scripts for other database types? I seem to remember saying we could move the migration towards using that model, seems like an appropriate time to do that.

melowe: the work to bundle and separate the create table lines is more hassle than the benefit . We'd need to remove the ddl module and perhaps integrate with the db migration, but then we'll need to redocument. The ddl module should never been added, but now it is we need to live with it for now.

melowe: Change added in response to comments. Good catch. We'll need to manually test as the unit tests don't cover the reading and execution of statements from ddls.

prd-fox (approved):

PR #684, opened 2019-04-10
Add vault runtime dependencies to remote enclave

chris-j-h:


melowe (changes requested):
Those dependencies drag too much baggage with them. 
Users can either bootstrap a shaded key vault jar  or/and have a shaded enclave that includes the dependencies.

chris-j-h:
Addressed by #685

PR #713, opened 2019-04-24
Add corsDomains config property

nicolae-leonte-go:
We need to decide whether we want to apply the DomainFilter to every resource or not.

### DIFF
--- a/config/src/main/java/com/quorum/tessera/config/ServerConfig.java
+++ b/config/src/main/java/com/quorum/tessera/config/ServerConfig.java
@@ -48,13 +48,17 @@ public class ServerConfig extends ConfigItem {
     @XmlElement
     private String serverAddress;
 
+    @XmlElement
+    private String corsDomains;
+
     public ServerConfig(final AppType app,
                         final boolean enabled,
                         final String serverAddress,
                         final CommunicationType communicationType,
                         final SslConfig sslConfig,
                         final InfluxConfig influxConfig,
-                        final String bindingAddress) {
+                        final String bindingAddress,
+                        final String corsDomains) {
         this.app = app;
         this.enabled = enabled;
         this.serverAddress = serverAddress;
@@ -62,6 +66,7 @@ public class ServerConfig extends ConfigItem {
         this.sslConfig = sslConfig;
         this.influxConfig = influxConfig;
         this.bindingAddress = bindingAddress;
+        this.corsDomains = corsDomains;
     }
 
     public ServerConfig() {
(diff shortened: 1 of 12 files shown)


server/jersey-server/src/main/java/com/quorum/tessera/server/JerseyServer.java
```diff
@@ -72,6 +72,8 @@ public void start() throws Exception {
         initParams.put("jersey.config.server.monitoring.statistics.enabled", "true");
         initParams.put("jersey.config.server.monitoring.enabled", "true");
         initParams.put("jersey.config.server.monitoring.statistics.mbeans.enabled", "true");
+        //custom properties
+        initParams.put("tessera.serverConfig", serverConfig);
```
melowe, line 76: No.. We either access config in usual way . This second route to config and high jacking the initParams should be avoided

config/src/main/java/com/quorum/tessera/config/ServerConfig.java
```diff
@@ -48,20 +48,25 @@
...
+
     public ServerConfig(final AppType app,
                         final boolean enabled,
                         final String serverAddress,
                         final CommunicationType communicationType,
                         final SslConfig sslConfig,
                         final InfluxConfig influxConfig,
-                        final String bindingAddress) {
+                        final String bindingAddress,
+                        final String corsDomains) {
```
melowe, line 61: Domains is a list of things not just a string

melowe, line 61: I'd also be tempted to not set this in constructor and default to null value. Will save having to change existing tests.

jaxrs-service/src/main/java/com/quorum/tessera/api/filter/DomainResponseFilter.java
```diff
@@ -31,4 +43,16 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
 
     }
 
+
+    @Context
+    public void setConfiguration(Configuration configuration) {
+        this.corsDomains = initCorsDomains(configuration);
+    }
+
+    Set<String> initCorsDomains(Configuration configuration) {
```
melowe, line 52: why is this protected?

config/src/main/java/com/quorum/tessera/config/ServerConfig.java
nicolae-leonte-go: yes, it is interpreted as a comma separated list of domains - see initCorsDomains.

melowe: exactly.

Issue #709, opened 2019-04-23
Tessera Cross Origin Issues while calling the rest API ( /storeraw )


muthu-param:
We are using tessera for execute private transaction's in Quorum. When invoking the third party (privateURL) API from the postman, we are getting expected output. but when we are calling the same API's from the another server we getting empty response with 200 status code( http://:9081/storeraw). This issue is happening due to the CORS. we checked with File:/// it's working.

How to enable CORS for tessera third party API's?

jiten-balla:
@fixanoid Can you please help us on this issue.

nicolae-leonte-go:
Setting CORS is not possible right now in Tessera. We'll add it to the backlog and let you know as soon as it is available.

For now please consider proxying the call via your server.

nicolae-leonte-go:
Alternatively if you are able to check out and rebuild tessera locally you can also try to add the response headers you need in com.quorum.tessera.thridparty.RawTransactionResource line 54:

Ex: 
        return Response.status(Status.OK)
                .type(APPLICATION_JSON)
                .header("Access-Control-Allow-Origin", "*")
                .entity(response)
                .build();

The tessera fix will most likely be similar to the geth --rpccorsdomain option but we may need a few days to implement that.

nicolae-leonte-go:
A CORS fix has been merged to master. You can get the latest docker image (snapshot build) using:
docker pull quorumengineering/tessera:latest

We intend to add configuration for the CORS domains - which will come in a later pull request.

PR #715, opened 2019-04-25
Cors domain filter

melowe:
Includes moving filter creation for common filters into jersey server module.


server/jersey-server/src/main/java/com/quorum/tessera/server/JerseyServer.java
```diff
@@ -76,7 +77,12 @@ public void start() throws Exception {
         final ResourceConfig config = ResourceConfig.forApplication(application);
 
         config.addProperties(initParams)
-                .register(MetricsResource.class);
+            .register(MetricsResource.class)
+            .register(LoggingFilter.class);
+
+        if (serverConfig.getCrossDomainConfig() != null) {
```
nicolae-leonte-go, line 83: Check if it is unix socket and don't add it... Then you can remove the check in the filter...

server/jersey-server/src/test/java/com/quorum/tessera/server/jaxrs/JerseyServerIT.java
```diff
@@ -39,12 +53,22 @@ public void onTearDown() throws Exception {
 
     @Test
     public void ping() {
+        
+        MultivaluedMap<String, Object> headers = new MultivaluedHashMap<>();
+        headers.add("Origin", "*.acme.com");
```
nicolae-leonte-go, line 58: wow - I was not expecting wildcards in the Origin header... I would always have expected that the Origin is exactly the site name from where the script attempting the CORS request has been downloaded.

nicolae-leonte-go (approved):

prd-fox:
Linked issue: https://github.com/jpmorganchase/tessera/issues/709

Issue #649, opened 2019-03-01
DOS attack on tessera network


ymlbright:
Tessera provided an API named `/partyinfo` on tessera P2P port to discover all peers in network. Attackers can inject malicious address to tessera node with certain tessera public key, that could case private transaction failed if it's `privateFor` contains that public key. What's worse, there is no way to recover from this violation unless all tessera nodes down, because of each tessera node will push all recipients what they known.

  In current implementation, Tessera is vulnerable if follow conditions satisfied:
    1. Attacker can access the tessera node P2P port
    2. Tessera node discovery is on
    3. Tessera Config `sslConfig.tls` isn't "STRICT", or is "STRICT" but trust module isn't "WHITELIST" or "CA"

  The reason of vulnerability is tessera choose remote tessera node randomly in it's recipients list[Source Code](https://github.com/jpmorganchase/tessera/blob/master/tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoServiceImpl.java#L112), and everyone can override recipients list if they can access API `/partyinfo`(if tessera node discovery is on)[Source Code](https://github.com/jpmorganchase/tessera/blob/master/tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoServiceImpl.java#L99).I have wrote an python PoC of this vulnerability, and I'm willing to share it with the contributor of the tessera project.

  In my view, tessera node discovery service should be tamperproof enough to handle this issue. Or if we have a way to update peer node list(keep it sync with config file in disk) without restart tessera.

cucrisis:
Hi @ymlbright  thanks sharing the finding with us.  We are triaging it right now.

PR #717, opened 2019-04-26
Add url validation when updating recipients.

melowe:
When party info is received that existing key = url recipient pairs match the updated ones. This prevents attempts for malicious parties to register themselves and effectively redirect transactions to them . This is a partial fix for issues raised in #649 . Additional node registration dialogue to ensure that newly joining nodes are able to decypt as expected to be completed in separate PR.

### DIFF
--- a/tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoRecipientUpdateCheck.java
+++ b/tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoRecipientUpdateCheck.java
@@ -1,0 +1,50 @@
+
+package com.quorum.tessera.node;
+
+import com.quorum.tessera.encryption.PublicKey;
+import com.quorum.tessera.node.model.PartyInfo;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PartyInfoRecipientUpdateCheck {
+    
+    private static final Logger LOGGER = LoggerFactory.getLogger(PartyInfoRecipientUpdateCheck.class);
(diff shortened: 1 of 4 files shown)


melowe:
Partially addresses concern raised in #649 . Prevent updating party info if recipient url is different from existing.

prd-fox (approved):

tessera-core/src/main/java/com/quorum/tessera/node/PartyInfoRecipientUpdateCheck.java
```diff
@@ -0,0 +1,50 @@
...
+            .stream()
+            .collect(Collectors.toMap(r -> r.getKey(), r -> r.getUrl()));
+
+        for(Map.Entry<PublicKey,String> e : newRecipientKeyUrlMap.entrySet()) {
+            PublicKey key = e.getKey();
+            if(existingRecipientKeyUrlMap.containsKey(key)) {
+                
+                String existingUrl = existingRecipientKeyUrlMap.get(key);
+                String newUrl = newRecipientKeyUrlMap.get(key);
+                if(!existingUrl.equalsIgnoreCase(newUrl)) {
```
nicolae-leonte-go, line 41: Does this mean that we do not allow a node to change it's IP? Especially when IPs are dynamically allocated it would be a reasonable feature... I would argue that this should be solved with a signature proving that the node sending the updated URL holds the private part of the key.

prd-fox: Correct. This is a stop gap to mitigate potentially sending payloads to an unintended party (see attached issue) until we do the message signing between peer updates.

PR #727, opened 2019-05-03
Issue 649 node validation

melowe:
Validate nodes sending party info by providing validation callback. When nodes attempt to send party info check they are trustworthy but sending them some data to be unencrypted.


jaxrs-service/src/main/java/com/quorum/tessera/p2p/PartyInfoResource.java
```diff
@@ -1,55 +1,143 @@
...
+            return isValid;
+
+        };
+
+        Predicate<Recipient> isSendingUrl = r -> r.getUrl().equalsIgnoreCase(url);
+        
+        //Validate caller and treat no valid certs as security issue.  
+        partyInfo.getRecipients()
+            .stream()
+            .filter(isValidRecipientKey.and(isSendingUrl))
```
prd-fox: We should check if the URL matches first before trying to make a call out
```suggestion
            .filter(isSendingUrl.and(isValidRecipientKey))
```

jaxrs-service/src/main/java/com/quorum/tessera/p2p/PartyInfoResource.java
```diff
@@ -1,55 +1,143 @@
...
+            .findFirst()
+            .orElseThrow(() -> new SecurityException("No key found for url " + url));
+
+        final Set<Recipient> recipients = partyInfo.getRecipients()
+            .stream()
+            .filter(isValidRecipientKey)
```
prd-fox, line 142: We should first filter on `isSendingUrl` to only try to check keys the sending node claimed to own, otherwise we will get a lot of expected misses.

prd-fox (changes requested):

jaxrs-service/src/main/java/com/quorum/tessera/p2p/PartyInfoResource.java
melowe: That’s the first check. The second is creating a filtered list of all recipients. If we don’t need to validate the recipients in party info then yes we can collect the initial filters

prd-fox: The first check only checks that at least one valid key exists. The second is creating a filtered list of all recipients, to try and remove any keys that are not owned by the node that sent us the PartyInfo. Any key that is in the PartyInfo that is not owned by that node will, of course, not pass the check and get filtered - but the node never claimed to have owned that key in the first place, so why check with them?

prd-fox: The first check only checks that at least one valid key exists. The second is creating a filtered list of all recipients, to try and remove any keys that are not owned by the node that sent us the PartyInfo. Any key that is in the PartyInfo that is not owned by that node will, of course, not pass the check and get filtered - but the node never claimed to have owned that key in the first place, so why check with them?

PR #751, opened 2019-05-17
Do not share party info

namtruong:
After node A successfully validate node B's attempt to send partyinfo, node A will not send his own partyinfo to node B to update.
This update and validation will be achieved later when node A tries to send his own partyinfo to node B.

For this change,
- `/partyinfo` will just respond with 200 instead of the updated partyinfo
- poller will poll parties without trying to update itself with results


Krish1979:
This fixes issues raised in #649.

Krish1979 (approved):

prd-fox:
Fixes https://github.com/jpmorganchase/tessera/issues/832

PR #771, opened 2019-05-29
Partyinfo module

melowe:
Moves the network information sync code into its own module. This will allow the core to not depend on the internals too much and give scope for other implementations.


prd-fox (approved):

prd-fox:
Part of https://github.com/jpmorganchase/tessera/issues/773

PR #788, opened 2019-06-12
Remove the deprecated config classes and update tests

### LINKED ISSUE #789: Remove deprecated server configuration
The server configuration has been deprecated and should be removed with the next major version.

This configuration style was used in versions 0.6 and 0.7, with automatic conversion happening for versions 0.8 and 0.9.

As of version 1.0, the conversion will not happen and only the new style will be accepted.

prd-fox:
Fixes https://github.com/jpmorganchase/tessera/issues/789

This removes the deprecated server configuration.

Since the next release is a major version change, this signals that there may be incompatible changes, of which this is one.

The Constellation to Tessera migration has been updated, as well as main application classes that converted between/validated the old style config.
Test files and sample files have been updated to remove references to the old configuration style.


config-migration/src/main/java/com/quorum/tessera/config/builder/ConfigBuilder.java
```diff
@@ -254,19 +254,26 @@ public Config build() {
...
         );
 
-        //TODO must add P2P and Q2T server configs. Maybe ThirdParty too - in disabled state.
-        //serverHostname, serverPort, 50521, CommunicationType.REST, sslConfig, null, null, null
-        final DeprecatedServerConfig serverConfig = new DeprecatedServerConfig();
-        serverConfig.setPort(serverPort);
-        serverConfig.setHostName(serverHostname);
-        serverConfig.setCommunicationType(CommunicationType.REST);
-        serverConfig.setSslConfig(sslConfig);
+        ///
```
chris-j-h: Can these be removed?

tests/acceptance-test/src/test/java/com/quorum/tessera/test/migration/config/ConfigMigrationSteps.java
```diff
@@ -103,41 +103,29 @@ public ConfigMigrationSteps() {
...
+            final JdbcConfig jdbcConfig = new JdbcConfig(null, null, "jdbc:h2:mem:tessera");
+
+            final ServerConfig q2tConfig = migratedConfig
+                .getServerConfigs()
+                .stream()
+                .filter(ServerConfig::isUnixSocket)
+                .findAny()
+                .get();
+
+            assertThat(migratedConfig.getP2PServerConfig().getServerAddress()).isEqualTo("http://127.0.0.1:9001");
```
chris-j-h, line 124: The migration of the sslConfig is not being asserted on anymore as the `serverConfigs` check has been removed.  This should be covered as part of the test.

chris-j-h (changes requested):
A few comments, looks good otherwise

prd-fox:
Updated @chris-j-h

chris-j-h (approved):

PR #791, opened 2019-06-13
Allow no Q2T servers to be specified in the configuration

### LINKED ISSUE #790: Do not restrict number of Q2T servers
The number of Q2T (Quorum-to-Tessera) servers is currently restricted to needing at least 1.

If we allowed no Q2T servers, then a Tessera node could act as a bootnode (a publicly known known node that all members could list in their initial peers, in order to bootstrap into the network), or as an archival node using the `alwaysSendTo` configuration parameters of a "sister node".

prd-fox:
Fixes https://github.com/jpmorganchase/tessera/issues/790

This change removes the requirement that at least 1 Q2T server must be specified.

This means a node can be used a boot node, or as a transaction store (backup node).

### DIFF
--- a/config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
+++ b/config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
@@ -1,5 +1,6 @@
 package com.quorum.tessera.config.constraints;
 
+import com.quorum.tessera.config.AppType;
 import com.quorum.tessera.config.ServerConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
```diff
@@ -14,39 +15,24 @@
...
+            .count();
 
         if (p2PEnabledConfigsCount != 1) {
-            LOGGER.debug("Only one P2P server must be configured and enabled.");
-            constraintContext.disableDefaultConstraintViolation();
-            constraintContext.buildConstraintViolationWithTemplate("Only one P2P server must be configured and enabled.")
-                .addConstraintViolation();
-            return false;
-        }
+            LOGGER.warn("Only one P2P server must be configured and enabled.");
```
chris-j-h: "Exactly one..." would be better wording

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
```diff
@@ -14,39 +15,24 @@
...
 
-        if (q2TEnabledConfigsCount == 0) {
-            LOGGER.debug("At least one Q2T server must be configured and enabled.");
             constraintContext.disableDefaultConstraintViolation();
-            constraintContext.buildConstraintViolationWithTemplate("At least one Q2T server must be configured and enabled.")
+            constraintContext
+                .buildConstraintViolationWithTemplate("Only one P2P server must be configured and enabled.")
```
chris-j-h: "Exactly one..." would be better wording

chris-j-h (changes requested):

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
```diff
@@ -13,40 +14,25 @@
...
+            .count();
 
         if (p2PEnabledConfigsCount != 1) {
-            LOGGER.debug("Only one P2P server must be configured and enabled.");
-            constraintContext.disableDefaultConstraintViolation();
-            constraintContext.buildConstraintViolationWithTemplate("Only one P2P server must be configured and enabled.")
-                .addConstraintViolation();
-            return false;
-        }
+            LOGGER.warn("Exactly one P2P server must be configured and enabled.");
```
nicolae-leonte-go, line 29: Considering people are having  multiple quorum nodes connected to a single tessera node - there may be scenarios where even the P2P is not required... :)

nicolae-leonte-go (approved):

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
prd-fox: This is true, but I think it is a little more complicated than that. There are other parts of the code where we make use of P2P API's (such as key validation) and I did not look deeply into what is exactly affected.

I agree that eventually we should relax this restriction, but it can be done separately. The Q2T one was an easy win for now.

chris-j-h (approved):

SatpalSandhu61:
Just a note that the docs should probably also be updated to indicate the Q2T server config is optional.

melowe:
Added comment to issue #790

melowe (changes requested):
See comment on issue #790

prd-fox:
Closing this as adding the flag will re-add all the code this PR removed, so it will be easier to start from scratch.

PR #810, opened 2019-07-01
Added extra debug info

### LINKED ISSUE #805: Extra logging for errors
The `DefaultExceptionMapper` is providing the stack trace at the debug level. 
This means that trying to diagnose issues can be troublesome without knowing where they occurred in the application.

Additionally, the `PartyInfoPoller` doesn't tell us the URL when a  general exception happens. This should be changed to be the same as the `ConnectException` clause.

We can either:
- increase the level they are logged at
- make it easier to run with the debug logging enabled

prd-fox:
This PR adds logging changes:

1. Standardises all the test logback configurations
2. Add the ability to switch the logging level of the root logger by specifying `-Dverbosity=DEBUG`, or whichever log level is desired. The default remains `INFO` level.
3. Add stack traces on the exception mappers at the `DEBUG` level. This will prove useful in the future as debugging remote logs has proven tricky at times.
4. Included log configurations only exist in shaded jars (or as close to as makes sense).
5. Added the date to the log output

Fixes https://github.com/jpmorganchase/tessera/issues/805


jaxrs-service/src/main/java/com/quorum/tessera/api/exception/EntityNotFoundExceptionMapper.java
```diff
@@ -9,13 +12,17 @@
...
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(EnclaveNotAvailableExceptionMapper.class);
+
     @Override
-    public Response toResponse(EntityNotFoundException e) {
+    public Response toResponse(final EntityNotFoundException ex) {
+        LOGGER.error("Entity not found: {}", ex.getMessage());
+        LOGGER.debug(null, ex);
+
         return Response.status(404)
```
SatpalSandhu61: Did you miss changing this one to `Response.status(Response.Status.NOT_FOUND)`

prd-fox: Yep, updated. Thanks!

SatpalSandhu61 (dismissed):

melowe (approved):

PR #817, opened 2019-07-13
Split rest endpoints and improve modularisation

melowe:
1. Move JAX-RS endpoints into distinct modules and don't use spring to instantiate JAX-RS objects. 
2. Move JPA focused code into distinct module to prevent circular dependency when trying to use DAOs in modules external to core/transaction
3. Remove GRPC modules and related code
4. Start phasing out loading spring managed objects via ServiceLocator, providing a ServiceFactory to limit access to ServiceLoader.


melowe:
1. Move JAX-RS endpoints into distinct modules and don't use spring to instantiate JAX-RS objects. 
2. Move JPA focused code into distinct module to prevent circular dependency when trying to use DAOs in modules external to core/transaction
3. Remove GRPC modules and related code
4. Start phasing out loading spring managed objects via ServiceLocator, providing a ServiceFactory to limit access to ServiceLoader.

prd-fox (approved):

PR #825, opened 2019-07-23
Add feature toggle configuration & configurable key validation

prd-fox:
Add feature toggle configuration for future features.
Either perform validation or disallow multiple URLs for a key, depending on the toggle activation.

This will allow breaking changes to be manually activated, until such a time as we want to do a release which enforces the breaking change to be active.


melowe (approved):

prd-fox:
Fixes https://github.com/jpmorganchase/tessera/issues/833

PR #845, opened 2019-08-06
Jasypt encryption

namtruong:
This PR enables jdbc password in configuration file to be obfuscated by using jasypt library for encryption/decryption.

- To enable the feature simply replace plain-text property with encrypted value and have it wrapped inside `ENC()` function. I.e. `"password": "ENC(ujMeokIQ9UFHSuBYetfRjQTpZASgaua3)"`.

- For details on how to encrypt a string using jasypt binary, see **Encrypt database password** section below.

Please note at the moment this encryption is only enabled on `jdbc.password`


Krish1979 (dismissed):

config/src/main/java/com/quorum/tessera/config/adapters/EncryptedStringAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(EncryptedStringAdapter.class);
+
+    private final PBEStringEncryptor encryptor = new StandardPBEStringEncryptor();
+
+    @Override
+    public String unmarshal(String textToDecrypt) {
+
+        if (PropertyValueEncryptionUtils.isEncryptedValue(textToDecrypt)) {
+
+            if (!((StandardPBEStringEncryptor) encryptor).isInitialized()) {
```
prd-fox: Should we not have the implementation as the member variable? No point casting every time for the sake of having an interface as the member variable.
I guess if you need convincing, then we do depend on this particular implementation, so that is what we should include, we don't depend on just the interface.

config/src/main/java/com/quorum/tessera/config/util/ConfigSecretReader.java
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+
+        LOGGER.warn("Not able to find or read any secret for decrypting sensitive values in config.");
+
+        return Optional.empty();
+    }
+
+    public static String readSecretFromConsole() {
+        System.out.println("Please enter the secret/password used to decrypt config value");
```
prd-fox: We should use `SystemAdapter.out()` when wanting to output directly to the console. This makes it easier for testing.

config/src/test/java/com/quorum/tessera/config/adapters/EncryptedStringAdapterTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    public void testUnMarshallWrongPassword() {
+
+        envVariables.clear(com.quorum.tessera.config.util.EnvironmentVariables.CONFIG_SECRET_PATH);
+
+        adapter = new EncryptedStringAdapter();
+
+        ByteArrayInputStream in = new ByteArrayInputStream(("bogus" + System.lineSeparator() + "bogus").getBytes());
+        System.setIn(in);
```
prd-fox: I don't think it matters that much, but we should reset `System.in` to the real value after the test is done.

config/src/test/java/com/quorum/tessera/config/adapters/EncryptedStringAdapterTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    public void testUnMarshallWithUserInputSecret() {
+
+        envVariables.clear(com.quorum.tessera.config.util.EnvironmentVariables.CONFIG_SECRET_PATH);
+
+        adapter = new EncryptedStringAdapter();
+
+        ByteArrayInputStream in = new ByteArrayInputStream(("quorum" + System.lineSeparator() + "quorum").getBytes());
+        System.setIn(in);
```
prd-fox: I don't think it matters that much, but we should reset `System.in` to the real value after the test is done.

config/src/test/java/com/quorum/tessera/config/util/ConfigSecretReaderTest.java
```diff
@@ -0,0 +1,80 @@
...
+        assertThat(ConfigSecretReader.readSecretFromFile()).isEmpty();
+    }
+
+    @Test
+    public void testReadFromConsole() {
+        ByteArrayInputStream in = new ByteArrayInputStream("password".getBytes());
+        System.setIn(in);
+        assertThat(ConfigSecretReader.readSecretFromConsole()).isEqualTo("password");
+
+        System.setIn(System.in);
```
prd-fox, line 61: I don't think it matters that much, but we should reset `System.in` to the real value after the test is done.

tests/acceptance-test/src/test/java/admin/cmd/StartupSteps.java
```diff
@@ -20,53 +20,58 @@ public StartupSteps() {
...
-            
-            assertThat(result.getOutput())
-                    .anyMatch(m -> m.startsWith("Config validation issue: keys.keyData[0].publicKey"));
+        Then(
+                "node returns error message and exits",
+                () -> {
+                    assertThat(results).hasSize(1);
+                    ExecutionResult result = results.get(0);
+                    assertThat(result.getExitCode()).isNotEqualTo(0);
+                    assertThat(result.getOutput()).hasSize(3);
```
prd-fox: If a new error is going to be output, we should add it to the list below to make sure it's the error we expect.

prd-fox (changes requested):

prd-fox:
Fixes https://github.com/jpmorganchase/tessera/issues/846

prd-fox (approved):

PR #918, opened 2019-11-11
Expose GET partyinfo API on ThirdParty server

namtruong:
This change enables peer/party network info of the node to be retrievable on ThirdParty server. The result now only contains list of public keys - data such as lastContact, url, etc... have been removed.

The same API on P2P server is still being kept just for testing purpose - however, this can potentially be deprecated at some point in the future. Coupled with #917, this api will no longer be accessible from a browser.

### DIFF
--- a/tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/PartyInfoResource.java
+++ b/tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/PartyInfoResource.java
@@ -44,7 +44,6 @@ public class PartyInfoResource {
 
         final String output =
                 Json.createObjectBuilder()
-                        .add("url", current.getUrl())
                         .add("keys", recipientBuilder.build())
                         .build()
                         .toString();
--- a/tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/PartyInfoResource.java
+++ b/tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/PartyInfoResource.java
@@ -1,0 +1,54 @@
+package com.quorum.tessera.thirdparty;
+
+import com.quorum.tessera.partyinfo.PartyInfoService;
+import com.quorum.tessera.partyinfo.model.PartyInfo;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+
+import javax.json.Json;
+import javax.json.JsonArrayBuilder;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import static java.util.Objects.requireNonNull;
+
+@Path("/partyinfo")
+public class PartyInfoResource {
+
+    private final PartyInfoService partyInfoService;
+
+    public PartyInfoResource(final PartyInfoService partyInfoService) {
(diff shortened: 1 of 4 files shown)


tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/PartyInfoResource.java
```diff
@@ -0,0 +1,54 @@
...
+                .map(
+                        recipient ->
+                                Json.createObjectBuilder()
+                                        .add("key", recipient.getKey().encodeToBase64())
+                                        .build())
+                .forEach(recipientBuilder::add);
+
+        final String output =
+                Json.createObjectBuilder()
+                        .add("url", current.getUrl())
```
nicolae-leonte-go: What is the URL for?

namtruong: removed

nicolae-leonte-go (approved):

tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/PartyInfoResource.java
```diff
@@ -0,0 +1,53 @@
...
+    public Response getPartyInfo() {
+
+        final PartyInfo current = this.partyInfoService.getPartyInfo();
+
+        final JsonArrayBuilder recipientBuilder = Json.createArrayBuilder();
+        current.getRecipients().stream()
+                .map(
+                        recipient ->
+                                Json.createObjectBuilder()
+                                        .add("key", recipient.getKey().encodeToBase64())
```
chris-j-h, line 41: Would it be better to return a simple array of public keys as opposed to an array of single field objects?  It would help to make the API easier to document

prd-fox: The main reason to do it like this is to future-proof it. We can add more details to the object without breaking users that already rely on this API.

chris-j-h: Sure - alternatively the endpoint could be `/partyinfo/keys` or something similarly specific, thereby leaving `/partyinfo` free for future use.  I'm not fussed either way but figured the question should be asked.

tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/PartyInfoResource.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private final PartyInfoService partyInfoService;
+
+    public PartyInfoResource(final PartyInfoService partyInfoService) {
+        this.partyInfoService = requireNonNull(partyInfoService, "partyInfoService must not be null");
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Fetch network/peer information", produces = "public list of peers/publickey mappings")
```
chris-j-h, line 30: `produces` should be a comma-separated list of content-types
http://docs.swagger.io/swagger-core/v1.5.X/apidocs/index.html?io/swagger/annotations/ApiOperation.html

namtruong: good spot - it seems that this is the common error across the whole jaxrs apis. Suggest having them clean up in a separate PR

PR #932, opened 2019-11-29
Fix "No value present" error when trying to use admin CLI

### LINKED ISSUE #933: "No value present" error when trying to use admin CLI
Admin CLI commands cannot be executed, e.g. 
```
tessera admin help
```
exits with error 
```
No value present
```

chris-j-h:
Resolves #933 

The removed filter line required that the command in use be a top level command.  The AdminCliAdapter is registered as a subcommand so this filter always returns 0.  

Removing this filter line ensures the command and subcommands are loaded correctly, and that the admin commands can be executed.


chris-j-h:
Replaced by #939

Issue #924, opened 2019-11-13
Add an endpoint to return list of public keys that are managed by enclave


namtruong:
This endpoint can be exposed via ThirdParty server and can be used by external clients to return a list of `privateFrom` to use to send transactions

PR #936, opened 2019-12-02
Api to return local public keys

namtruong:
Api to return a list of `privateFrom` and to be called via ThirdParty server.
See #924 for details


melowe (approved):

tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/KeyResource.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public KeyResource(ConfigService configService) {
+        this.configService = configService;
+    }
+
+    @GET
+    @ApiOperation(value = "Fetch local public keys managed by the enclave")
+    @ApiResponses({@ApiResponse(code = 200, message = "Managed public keys")})
+    public Response getPublicKeys() {
+        Set<PublicKey> publicKeys = configService.getPublicKeys();
```
nicolae-leonte-go: I wonder if it is possible to add an extra token with each public key that proves that the specific tessera node answering the request also holds the private part of each key (similar to how it was done in party info).
It would prevent a malicious tessera masquerading as a different node.

Krish1979: It would still not be able to receive transactions for the key if the node validation is enabled.

nicolae-leonte-go: But it may cause a third party to connect to the wrong tessera to send an externally signed private transaction (storeraw). I'm not saying this is opening up a big security hole (P2P already validates party info) - just that a malicious node can now do some phishing (especially in a large network).

I also understand that this is meant for cakeshop (and not for 3rd party - external tx signing - users).

Krish1979 (approved):

PR #947, opened 2019-12-20
Adding support for AWS Secrets Manager 

Emi14:
Quorum documentation PR - https://github.com/jpmorganchase/quorum/pull/919

PR #948, opened 2019-12-24
Generalised key vault config

melowe:
Add generalised key vault config that contains a map of generalised properties.

Add validator for key vault config that is aware of vault type and is able to apply validations for
that given type. Deprecate the Hashicorp and Azurekey vault config. Service factories use KeyVaultConfig rather than its deprecated implementations. 

This provides room to add other key vaults such as the AWS one in #947.


Krish1979 (dismissed):

Krish1979 (approved):

chris-j-h:
Resolves #959

Issue #960, opened 2020-01-08
Partyinfo polling - configurable option


Krish1979:
Currently Tessera polls peers every 2 seconds for all its peers which becomes more laborious with increased number of peers in the network.. So increase this to every 20 seconds as default and also allow for the value to be configurable by users. The timeout should be amended to be a factor within the polling period (say 3/4th) and this way if a connection is down the downed thread could be reassigned to another connection.

PR #969, opened 2020-01-13
Configurable p2p polling interval

namtruong:
Introduce new server config item called syncInterval - this will be the rate at which partyinfo rest calls are made.
Default value is 5000 milliseconds.
The timeout of individual requests also will be configured to be 3/4 of the interval if this syncInterval value is set by user

Details at #960


config/src/main/java/com/quorum/tessera/config/ServerConfig.java
```diff
@@ -53,7 +46,10 @@
     @XmlElement(name = "cors")
     private CrossDomainConfig crossDomainConfig;
 
-    public ServerConfig(final AppType app,
+    @XmlElement private Integer syncInterval;
```
prd-fox: If you change this to be `@XmlElement private int syncInterval = 5000;` and add the @Min(value = 1L) validation annotation, then it should be set up to avoid nulls or negative values (and reduce checks elsewhere).

What do you think?

melowe: don't use primitive for config values please. But sync interval doesn't belong on a server config. lets not waste time commenting here and take conversions offline or go in and make changes directly.

prd-fox: Sure, can avoid using a primitive. We can still set the default value and add the validation rule for the same effect (wherever it may end up).

melowe (dismissed):

melowe (approved):

Krish1979:
Fixes #960

PR #971, opened 2020-01-14
Filter requests based on peers/whitelist . Deprecate useWhiteList 

melowe:
Remove deprecated constructor for config object use default construct or and setters to populate values. Remove catch all block and disabling filter as this doesn't doesn't make sense. Deprecate useWhiteList and use disableAutoDiscovery in its place. For time using when useWhiteList is set assume that this means auto discovery is enabled.


tessera-jaxrs/common-jaxrs/src/test/java/com/quorum/tessera/api/filter/IPWhitelistFilterTest.java
```diff
@@ -4,154 +4,172 @@
...
-        this.filter = new IPWhitelistFilter(configService);
+    @After
+    public void onTearDown() {
+        verifyNoMoreInteractions(containerRequestContext);
+        verifyNoMoreInteractions(configService);
     }
 
     @Test
-    public void disabledFilterAllowsAllRequests() {
+    public void autoDiscoveryEnabledAndHostInPeersList() {
```
chris-j-h:
```suggestion
    public void autoDiscoveryDisabledAndHostInPeersList() {
```

tessera-jaxrs/common-jaxrs/src/test/java/com/quorum/tessera/api/filter/IPWhitelistFilterTest.java
```diff
@@ -4,154 +4,172 @@
...
+        verify(containerRequestContext).getUriInfo();
 
-        verifyZeroInteractions(request);
-        verifyZeroInteractions(ctx);
+        verifyNoMoreInteractions(request);
     }
 
     @Test
-    public void hostNotInWhitelistGetsRejected() {
+    public void autoDiscoveryEnabledAndHostNotInPeersList() {
```
chris-j-h:
```suggestion
    public void autoDiscoveryDisabledAndHostNotInPeersList() {
```

tessera-jaxrs/common-jaxrs/src/main/java/com/quorum/tessera/api/filter/IPWhitelistFilter.java
```diff
@@ -1,27 +1,28 @@
...
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
- * Applies a filter to all endpoints that only allows certain IP address and ghost names to get access to the HTTP
- * endpoints
+ * If peer discovery is enable, applies a filter to all endpoints that only allows certain IP address
```
chris-j-h:
```suggestion
 * If peer discovery is enabled, applies a filter to all endpoints that only allows certain IP address
```

chris-j-h (dismissed):

melowe:
**Initial issue description**

> Currently when discovery is OFF, node A will not discover new nodes in the network and fail any transactions to new nodes say 'B', however node A will accept transactions from node B if node B discovered node A through other nodes in network (Node B has discovery=ON) which shouldn't be the case. When discovery is OFF, it should enable whitelist to only connect to the peers present in the config file and thereby connection to new nodes (waiting to be discovered) are terminated even before any API communication is tried.

namtruong:
Closing this PR as the requirements have changed. IPWhiteList could still prove to be useful to prevent transactions being pushed from unknown nodes.

Issue #1000, opened 2020-02-20
Consider reverting "Enforce whitelist if discovery disabled" to 0.10.2


yunhochung:
Hello.

I'm building up the network on AWS. 

Inbound; NLB(with EIP) - Nodes(on Private N/W)
Outbound: Nodes(on Private N/W) - NAT Gateway

I let some nodes on other cloud/on-premise know that my tessera's IP is two for inbound, outbound.
And most enterprise company's firewall does not open for any source IP except Web Server.
So, currently my consortium doesn't "Discovery" and "Whitelist" function.

If discovery enabled and whiltelist disabled, all network including tessera have to open the firewall  for any IP because connecting to some nodes that not including peers list.
If discovery disabled and whilelist enabled, some tesseras have a problem like 401 Unauthorized because connecting to NAT Gateway IP.

On 0.10.2 version, the firewall policy controls the inbound traffic for tessera.
On 0.10.3 version, discovery enabled and whitelist disabled but connection timeout occurred because connecting to some nodes that not including peers list.(not open firewall)

Why don't you leave the control to the user?

Thank you.

namtruong:
Hi @yunhochung , thanks for raising this.

The change to enforce the use of whitelist is to prevent a security incident where there is a mixed setup of auto-discovery feature in a Tessera network; however, this indeed does have an impact on some other use cases - one of which you just described.

We will review this internally and revert the change - hopefully release a patch next week.

Many thanks,
Nam

PR #1002, opened 2020-02-24
No longer enforcing whitelist feature when discovery is disabled

namtruong:
The change to enforce the use of whitelist is to prevent a security incident where there is a mixed setup of auto-discovery feature in a Tessera network; however, this change does have an impact on some other use cases.

This PR is to revert the change.

We will re-evaluate the usage of "whitelist" feature


namtruong:
Fixes #1000

nicolae-leonte-go (approved):

PR #1057, opened 2020-05-04
Use unique payloads for partyinfo validation

chris-j-h:
This maintains security if the sender has multiple keys, as it prevents the result of one validation check being reused for the validation check of another key.


SatpalSandhu61 (approved):

Krish1979:
Fixes #1058

PR #1084, opened 2020-06-11
Allow no Q2T servers to be specified on a bootstrap node

### LINKED ISSUE #790: Do not restrict number of Q2T servers

SatpalSandhu61:
Fixes #790 
Add new config item "bootstrapNode" which facilitates setup of a bootstrap node with no Q2T serverConfig entries.

### DIFF
--- a/config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
+++ b/config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
@@ -1,45 +1,57 @@
 package com.quorum.tessera.config.constraints;
 
 import com.quorum.tessera.config.AppType;
+import com.quorum.tessera.config.Config;
 import com.quorum.tessera.config.ServerConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
-import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-public class ServerConfigsValidator implements ConstraintValidator<ValidServerConfigs, List<ServerConfig>> {
+public class ServerConfigsValidator implements ConstraintValidator<ValidServerConfigs, Config> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ServerConfigsValidator.class);
 
     @Override
-    public boolean isValid(List<ServerConfig> serverConfigs, ConstraintValidatorContext constraintContext) {
-        if(serverConfigs == null) {
+    public boolean isValid(Config config, ConstraintValidatorContext constraintContext) {
+        if (config == null || config.getServerConfigs() == null || config.getServerConfigs().isEmpty()) {
             return true;
         }
 
-        final Map<AppType,Integer> counts = serverConfigs.stream()
-            .collect(Collectors.toMap(ServerConfig::getApp,v -> 1,(l,r) -> l + 1));
+        final Map<AppType, Integer> counts =
+                config.getServerConfigs().stream()
+                        .collect(Collectors.toMap(ServerConfig::getApp, v -> 1, (l, r) -> l + 1));
 
-        final int p2PEnabledConfigsCount = counts.getOrDefault(AppType.P2P,0);
-        final int q2TEnabledConfigsCount = counts.getOrDefault(AppType.Q2T,0);
+        final int p2PEnabledConfigsCount = counts.getOrDefault(AppType.P2P, 0);
+        final int q2TEnabledConfigsCount = counts.getOrDefault(AppType.Q2T, 0);
 
         if (p2PEnabledConfigsCount != 1) {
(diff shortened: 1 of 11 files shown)


Krish1979:
LFTM

Krish1979 (dismissed):

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
```diff
@@ -1,45 +1,57 @@
...
-public class ServerConfigsValidator implements ConstraintValidator<ValidServerConfigs, List<ServerConfig>> {
+public class ServerConfigsValidator implements ConstraintValidator<ValidServerConfigs, Config> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ServerConfigsValidator.class);
 
     @Override
-    public boolean isValid(List<ServerConfig> serverConfigs, ConstraintValidatorContext constraintContext) {
-        if(serverConfigs == null) {
+    public boolean isValid(Config config, ConstraintValidatorContext constraintContext) {
+        if (config == null || config.getServerConfigs() == null || config.getServerConfigs().isEmpty()) {
```
prd-fox: I don't think this should return true in the last case (`config.getServerConfigs().isEmpty()`). This allows the application to start up when no servers are defined, when we mandate that at least one P2P server should be there. 

The reason it is okay to have `config.getServerConfigs() == null` is because that is checked by the `@ValidEitherServerConfigsOrServer` annotation. Maybe we should add a comment here saying that since it is rather indirect.

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
```diff
@@ -1,45 +1,57 @@
...
+        final int p2PEnabledConfigsCount = counts.getOrDefault(AppType.P2P, 0);
+        final int q2TEnabledConfigsCount = counts.getOrDefault(AppType.Q2T, 0);
 
         if (p2PEnabledConfigsCount != 1) {
             LOGGER.debug("Only one P2P server must be configured and enabled.");
             constraintContext.disableDefaultConstraintViolation();
-            constraintContext.buildConstraintViolationWithTemplate("Only one P2P server must be configured and enabled.")
-                .addConstraintViolation();
+            constraintContext
+                    .buildConstraintViolationWithTemplate("Only one P2P server must be configured and enabled.")
```
prd-fox: Should we change this message to from `Only` to `Exactly`? It reads a bit funny if no P2P servers are defined, since it implies multiple have been.

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
```diff
@@ -1,45 +1,57 @@
...
+                        .addConstraintViolation();
+                return false;
+            }
+        } else if (q2TEnabledConfigsCount == 0) {
             LOGGER.debug("At least one Q2T server must be configured and enabled.");
             constraintContext.disableDefaultConstraintViolation();
-            constraintContext.buildConstraintViolationWithTemplate("At least one Q2T server must be configured and enabled.")
-                .addConstraintViolation();
+            constraintContext
+                    .buildConstraintViolationWithTemplate("At least one Q2T server must be configured and enabled.")
```
prd-fox: `and enabled` I think can be removed since we removed the ability for a server to be in a disabled state.

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
prd-fox: `and enabled` I think can be removed since we removed the ability for a server to be in a disabled state.

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
prd-fox: I think we should add `or bootstrap mode enabled` to hint the user as to what they should do if they meant to have no Q2T servers.

SatpalSandhu61: Good points

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
chris-j-h: Same for the log msg on line 32

config/src/main/java/com/quorum/tessera/config/constraints/ServerConfigsValidator.java
chris-j-h: Same for log msg on line 50

prd-fox (approved):

PR #1085, opened 2020-06-14
Replace acceptance test setup with docker image

### LINKED ISSUE #1070: Use containerised acceptance-tests in ghactions workflows
[The quorum-acceptance-tests have been containerised](https://hub.docker.com/r/quorumengineering/acctests).  This should be used in the ghactions acceptance tests jobs.

This will simplify the workflow configuration, and make it easier to perform more stringent testing (e.g. with other network configurations).

This will require a code change in `quorum-acceptance-tests` so that the `tessera` image being used can be configured.  This will allow for a local image to be built from the branch under test and used in the `quorum-acceptance-test` container.

prd-fox:
The acceptance tests now has a docker image that handles the installation of all the deps. By using this, we can avoid needing to set them up ourselves, and will be auto-updated as the acceptance tests update.

It expects the tessera to use to be under the `latest` tag, so first we build a docker image locally under that tag using the newly built tessera.

Fixes #1070

### DIFF
--- a/.github/workflows/DockerfileTest
+++ b/.github/workflows/DockerfileTest
@@ -1,0 +1,8 @@
+# Create docker image with only distribution jar
+
+FROM adoptopenjdk/openjdk11:alpine
+
+ARG JAR_FILE
+COPY ${JAR_FILE} /tessera/tessera-app.jar
+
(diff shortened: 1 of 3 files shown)


chris-j-h:
Fixes #1070

chris-j-h (approved):

Issue #773, opened 2019-05-30
Support stateful network information


prd-fox:
The current model of the Tessera network is that all nodes information is kept indefinitely (until a restart). This means that over time, the network information can be filled up with old and dead nodes that no longer exist.

This is exacerbated by the fact that this old information is propagated to any new node that joins the network.

---

This information should be detected and removed, so that the list of network participants is restricted (within reason) to the real set of live participants.

PR #1107, opened 2020-07-28
Remove inactive nodes from party info. 

melowe:
Add cache to store nodes that are offline to supress party info updates from other parties including the offline node's info. Assume that jaxrs processing exception is connectivity related and return 502 status when nodes are unable to push transactions to its peers.

When asynchronously publishing party info remove nodes if there are any exceptions. Errors are logged as warning (existing behaviour) , node is removed from party info but no error status is returned to caller.


melowe:
Fixes #773

PR #1122, opened 2020-08-12
Introduce and use dedicated NodeInfo publisher

prd-fox:
Removes P2PClient whose only function left was to publish PartyInfo payloads in encoded form.

Introduce a NodeInfo publisher which will handle the conversion to PartyInfo and its encoded form, leaving users of the publisher free from the details.

---

Both classes are similar in function, you could class the P2PClient and being renamed to NodeInfoPublisher, with the NodeInfoPublisher taking on some extra functionality.


melowe:
NodeInfo is a temporary class that helps us migrate from PartyInfo. We're in the process of investigating server side events for the p2p dialogues to address the issues with various pollers and theoretical race conditions we're coding for.

PR #1216, opened 2021-01-19
Support privacy group in Tessera

namtruong:
Introduce the concept of privacy group in Tessera, add support for operations related to privacy groups such as managing privacy group data and handling transactions that are intended for a privacy group.

New APIs created include

- `/createPrivacyGroup`
- `/retrievePrivacyGroup`
- `/findPrivacyGroup`
- `/deletePrivacyGroup`

A new configuration flag is introduced to support current Besu users running Orion as its privacy manager `{ "mode": "orion"}` (default value is `"tessera"`). Once enabled, Tessera will alter its behaviour to accept requests from Besu and return responses that are complying to the existing Besu<>Orion models.

PR #1221, opened 2021-01-25
Privacy group database table and ddl

namtruong:
A part of #1216 - this is more specific to db operations and create table ddls


prd-fox (approved):

ricardolyn (commented):
👍

PR #1234, opened 2021-02-01
Update API doc Github Action Workflow

### LINKED ISSUE #40: Version API docs
The Tessera API is versioned (see https://github.com/ConsenSys/tessera/pull/1206) but it appears only 'master(latest' can be selected as the version on the API reference (https://consensys.github.io/tessera/)

NicolasMassart:
## Changes

This PR adds a new workflow providing the versioning of the API docs:

- latest is generated for all push on master
- release is generated only by push on tags on master (when doing a release)

Latest is always updated and release yaml file is only created on release.

Latest version either contains the commit sha if it's a regular push on master and spec title will display `Tessera API reference (0ecbcf83)` for example or the version if it's a release `Tessera API reference (tessera-21.02.4)` for example.

The changes are automatically pushed to the gh-pages branch using github-pages-deploy-action. No more PR required as anyway the process was only to merge the PR and no changes were made.

These changes also make the API spec file and versions dropdown update even if no changes append in the API.
Previously, the spec was updated only if changes were made in it.
Having it published each time is expected and made to ensure users will always find an API doc related to the version they use, even if there's no change between the specs.

## Testing the changes

This PR doesn't run this workflow as it's only on master branch but you can look at https://github.com/NicolasMassart/tessera/actions/runs/528902730 for latest and https://github.com/NicolasMassart/tessera/actions/runs/528926041 for release runs.

See the final result at https://nicolasmassart.github.io/tessera/

fixes consensys/doc.tessera#40


ricardolyn (commented):
LGTM 👍

chris-j-h (approved):

PR #1236, opened 2021-02-02
Privacy group resources

namtruong:
A part of #1216.

This PR is specific to privacy group resources on q2t and p2p app to expose apis for privacy group operations


tessera-jaxrs/common-jaxrs/src/main/java/com/quorum/tessera/api/PrivacyGroupRequest.java
```diff
@@ -0,0 +1,65 @@
...
+    @Schema(description = "public key identifying the sender of the request", type = "string", format = "base64")
+    private String from;
+
+    @Schema(description = "name of the privacy group", type = "string")
+    private String name;
+
+    @Schema(description = "description of the privacy group", type = "string")
+    private String description;
+
+    private String seed;
```
ricardolyn, line 22: should have a `@Schema` tag with the metadata?

tessera-jaxrs/transaction-jaxrs/src/test/java/com/quorum/tessera/q2t/MockPrivacyGroupManager.java
```diff
@@ -0,0 +1,44 @@
+package com.quorum.tessera.q2t;
+
+import com.quorum.tessera.enclave.PrivacyGroup;
+import com.quorum.tessera.encryption.PublicKey;
+import com.quorum.tessera.privacygroup.PrivacyGroupManager;
+
+import java.util.List;
+
+public class MockPrivacyGroupManager implements PrivacyGroupManager {
```
ricardolyn, line 9: non-blocking/question: why do we declare a Mock (seems more like a stub) instead of using Mockito to create this mock at runtime?

namtruong: this is due to how we use static create method to instantiate the managers using `ServiceLoaderUtil` i.e. `return ServiceLoaderUtil.load(PrivacyGroupManager.class)`. Using the new `Mockito.mockStatic()` would be an alternative. 
However, these patterns will change when we migrate to jigsaw

tessera-jaxrs/common-jaxrs/src/main/java/com/quorum/tessera/api/PrivacyGroupRequest.java
namtruong: I'm not sure if this property should be promoted to the users. Ideally it should be left empty and a random seed would be generated

tessera-jaxrs/transaction-jaxrs/src/test/java/com/quorum/tessera/q2t/MockPrivacyGroupManager.java
ricardolyn: 👍

tessera-jaxrs/common-jaxrs/src/main/java/com/quorum/tessera/api/PrivacyGroupRequest.java
ricardolyn: 👍

Krish1979 (approved):

PR #1251, opened 2021-03-18
Build path to key using workdir in Orion config.

melowe:
Build path to key files using workdir defined in orion config. remove verbose flag as doesn't do anything. Add MigrateConfigCommandTest to make it easier to verify and to test key paths.


Krish1979 (approved):

macfarla (commented):
LGTM

PR #1270, opened 2021-04-22
add the receive method used by Besu back in

pinges:
This method has been removed in 710eff69d on the 5th of Feb

Signed-off-by: Stefan Pingel <stefan.pingel@consensys.net>


Krish1979:
Besu on 'goquorum' mode should use /transaction/{hash} connecting to Tessera ‘normal’ mode

PR #1273, opened 2021-04-28
Resident privacy group and validations

namtruong:
- Introduce feature config `enableMultiplePrivateStates` to accommodate Quorum handling multiple private states
- Introduce `RESIDENT` privacy group type
- Allow resident groups and its member to be configured via config file
- Resident groups configured will be validated and created upon application launch.
- Introduce `groups` API to return a list of privacy groups of a specific type. Usage `groups/resident`
- A member cannot belong to more than 1 resident group, nor switch between groups.
- If user decides to switch on mps feature, all keys configured must belong to a resident group.


nicolae-leonte-go (approved):
LGTM

PR #1320, opened 2021-07-26
metrics formatting: preserve trailing whitespace

macfarla:
The newline character is required at the end of this string. 
Any other whitespace is ignored as per https://prometheus.io/docs/instrumenting/exposition_formats/ so no need to trim

I have tested this locally and get the following response from 
` curl localhost:8581/metrics`
Before ie without this change:
```

tessera_P2P_GET_getPartyInfo_MinTime_ms 0
tessera_P2P_GET_getPartyInfo_MaxTime_ms 0
...
tessera_THIRD_PARTY_GET_getVersion_RequestRate_requestsPerSeconds 0.0
tessera_THIRD_PARTY_GET_getVersion_RequestCount 0%
```
^ no new line at end of string

After ie with this change:
```
tessera_P2P_GET_getPartyInfo_MinTime_ms 0
tessera_P2P_GET_getPartyInfo_MaxTime_ms 0
...
tessera_THIRD_PARTY_GET_getVersion_RequestRate_requestsPerSeconds 0.0
tessera_THIRD_PARTY_GET_getVersion_RequestCount 0
```
so there is a newline at the end. Prometheus serves up the list of metrics.

### DIFF
--- a/server/jersey-server/src/main/java/com/quorum/tessera/server/monitoring/MetricsResource.java
+++ b/server/jersey-server/src/main/java/com/quorum/tessera/server/monitoring/MetricsResource.java
@@ -39,7 +39,7 @@ public class MetricsResource {
 
     return Response.status(Response.Status.OK)
         .header(HttpHeaders.CONTENT_TYPE, TEXT_PLAIN)
-        .entity(formattedMetrics.toString().trim())
+        .entity(formattedMetrics.toString())
         .build();
   }
 }
--- a/server/jersey-server/src/main/java/com/quorum/tessera/server/monitoring/MetricsResource.java
+++ b/server/jersey-server/src/main/java/com/quorum/tessera/server/monitoring/MetricsResource.java
@@ -39,7 +39,7 @@ public class MetricsResource {
 
     return Response.status(Response.Status.OK)
         .header(HttpHeaders.CONTENT_TYPE, TEXT_PLAIN)
-        .entity(formattedMetrics.append("\n").toString())
+        .entity(formattedMetrics.toString())
         .build();
   }
 }
--- a/server/jersey-server/src/main/java/com/quorum/tessera/server/monitoring/MetricsResource.java
+++ b/server/jersey-server/src/main/java/com/quorum/tessera/server/monitoring/MetricsResource.java
@@ -34,12 +34,12 @@ public class MetricsResource {
       List<MBeanMetric> metrics = metricsEnquirer.getMBeanMetrics(type);
       PrometheusProtocolFormatter formatter = new PrometheusProtocolFormatter();
 
-      formattedMetrics.append(formatter.format(metrics, type)).append("\n");
+      formattedMetrics.append(formatter.format(metrics, type));
     }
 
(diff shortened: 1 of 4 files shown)


server/jersey-server/src/main/java/com/quorum/tessera/server/monitoring/MetricsResource.java
```diff
@@ -39,7 +39,7 @@ public Response getMetrics() {
 
     return Response.status(Response.Status.OK)
         .header(HttpHeaders.CONTENT_TYPE, TEXT_PLAIN)
-        .entity(formattedMetrics.toString().trim())
+        .entity(formattedMetrics.toString())
```
chris-j-h, line 42: Instead of ensuring there's a single new line at the end of the response, this can actually result in multiple new lines (1 for each `AppType` that isn't in use).  Empty lines are fine but we can address this and simplify some things by removing the trim in `PrometheusProtocolFormatter`, which also means we can get rid of the `append("\n")` in `MetricsResource` i.e.

```diff
- formattedMetrics.append(formatter.format(metrics, type)).append("\n");
+ formattedMetrics.append(formatter.format(metrics, type)); 
```

chris-j-h (commented):
It's worth mentioning that Prometheus does correctly ingest and parse metrics without the trailing newline (checked with v2.28.1).  However, as the Prometheus spec does state that *"The last line must end with a line feed character"*, third party tools may enforce this (as seen in the linked issue), and future versions of Prometheus may not correctly handle no trailing newline.

macfarla:
> It's worth mentioning that Prometheus does correctly ingest and parse metrics without the trailing newline (checked with v2.28.1). However, as the Prometheus spec does state that _"The last line must end with a line feed character"_, third party tools may enforce this (as seen in the linked issue), and future versions of Prometheus may not correctly handle no trailing newline.

This is true, Prometheus parses it fine, but other packages eg Telegraf do not. This is what the user reported.

server/jersey-server/src/main/java/com/quorum/tessera/server/monitoring/MetricsResource.java
macfarla: @chris-j-h - made the change you suggested. Are you able to help with the build issues?

macfarla:
```
[@octokit/rest] `const Octokit = require("@octokit/rest")` is deprecated. Use `const { Octokit } = require("@octokit/rest")` instead
Error:  Missing Slack Incoming Webhooks URL.
      Please configure "SLACK_WEBHOOK" as environment variable or
      specify the key called "url" in "with" section.
```

chris-j-h: I've disabled the slack notification step for the time being so this can be merged while we debug - i also moved the newline formatting into the PrometheusProtocolFormatter so that we can add some unit tests

melowe (approved):

PR #1329, opened 2021-09-07
Get mandatory api

namtruong:
Add GET `/transaction/{hash}/mandatory` to retrieve the list of mandatory recipients for a given transaction hash.

This is to support contract extension operation


Krish1979 (approved):

Krish1979:
LGTM

PR #1336, opened 2021-09-24
Fix invalid jpql queries

melowe:


namtruong:
One possibility for failing is that the eclipselink jpql validation doesn't play well with `coalesce` and/or `case...when` function. This is not considered priority at the moment and can be looked at later

PR #1347, opened 2021-11-24
Decouple payload encoder and add CBOR as an encoding mechanism

namtruong:
This PR adds another type of payload encoding (CBOR) as the preferred codec with the view of replacing the legacy binary encoding mechanism. This is to make the `EncodedPayload` more extensible and also an enabler to accommodate any interoperability requirements in the future. Changes include

- Introduce `EncodedPayloadCodec` and store in an extra database column. Purpose of this is to maintain backward compatibility.
- Add `EntityListeners` that handle data encoding/decoding from/to database
- Heavy refactoring to decouple the payload encoder and perform lookup to work out the relevant encoder to use based on codec
- Mapping of api versions and preferred codec - publisher to encode and publish payload in preferred format depending on supportedApiVersions of the remote node
- Add CBOREncoder
- Refactoring unit tests

For now the scope of the change is for pushing payload between Tessera peer-to-peer

Part of the change was initially done in PR #1335 by @melowe


enclave/enclave-api/src/main/java/com/quorum/tessera/enclave/EncodedPayloadCodec.java
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  public static EncodedPayloadCodec current() {
+    return CBOR;
+  }
+
+  public static EncodedPayloadCodec getPreferredCodec(Set<String> versions) {
+    return Stream.of(EncodedPayloadCodec.values())
+        .filter(codec -> versions.contains(codec.getMinimumSupportedVersion()))
+        .findFirst()
```
prd-fox, line 36: Should there be some sorting here so we always pick the highest API version? It is probably goes for version 5 first most of the time, but don't think it would be guarenteed.

tessera-dist/build.gradle
```diff
@@ -30,8 +30,8 @@ configurations.all {
   exclude group: "io.swagger.core.v3", module: "swagger-parser"
   exclude group: "io.swagger.core.v3", module: "swagger-jaxrs2-jakarta"
   exclude group: "com.fasterxml.jackson.datatype"
-  exclude group: "com.fasterxml.jackson.core"
-  exclude group: "com.fasterxml.jackson.dataformat"
+  //  exclude group: "com.fasterxml.jackson.core"
+  //  exclude group: "com.fasterxml.jackson.dataformat"
```
prd-fox: Delete

tessera-recover/src/main/java/com/quorum/tessera/recovery/internal/RecoveryImpl.java
```diff
@@ -127,9 +122,8 @@ public RecoveryResult sync() {
                   .filter(
                       t -> {
                         payloadCount.incrementAndGet();
-                        byte[] payload = t.getPayload();
+                        EncodedPayload encodedPayload = t.getEncodedPayload();
```
prd-fox, line 125: Should this happen inside the `try/catch`? If an error occurs reading from the database or decoding the data, the exception won't be caught - although I appreciate the db read error already wasn't handled.

enclave/enclave-api/src/main/java/com/quorum/tessera/enclave/EncodedPayloadCodec.java
namtruong: The order of enum values() method is guaranteed - being the order they are declared. However it's true that this might not be the best practice - I'm adding a sort in case this declaration is changed in the future.

tessera-recover/src/main/java/com/quorum/tessera/recovery/internal/RecoveryImpl.java
namtruong: This is at sync stage and in this block of code the staging tnxs have been retrieved - so we probably don't expect such exception at this point. If error occurs reading or decoding data it should be thrown at dao

Krish1979:
LGTM

Krish1979 (approved):

prd-fox (approved):

chris-j-h (approved):

PR #1356, opened 2021-12-06
Added CLI options to output server URIs to file.

mark-terry:
Added a CLI option to write server URIs to a specified directory. The use case for this change is to enable an external process to locate running Tessera servers, without requiring foreknowledge of the Tessera configuration.


mark-terry:
Still needs tests.

cli/config-cli/src/main/java/com/quorum/tessera/config/cli/TesseraCommand.java
```diff
@@ -71,6 +71,16 @@ private TesseraCommand(final KeyPasswordResolver keyPasswordResolver, Validator
       description = "Start Tessera in recovery mode")
   private boolean recover;
 
+  @CommandLine.Option(
+      names = {"--outputServerURIs"},
```
pinges: Would it not be enough to just set the output path? Setting the path could mean, yes I do want the path, and here is the path?

pinges (commented):
just not sure whether we need two CLI options to do that ...
Otherwise LGTM.

tessera-context/src/main/java/com/quorum/tessera/context/RuntimeContext.java
```diff
@@ -37,6 +37,10 @@
 
   boolean isOrionMode();
 
+  boolean getOutputServerURIs();
```
namtruong: These 2 properties (arguably only 1 is necessary) might not be needed in the RuntimeContext

tessera-dist/src/main/java/com/quorum/tessera/launcher/Launcher.java
```diff
@@ -127,10 +150,51 @@ public void launchServer(Config config) throws Exception {
...
   public abstract void launchServer(Config config) throws Exception;
 
-  public static Launcher create(boolean isRecoveryMode) {
+  public static Launcher create(final boolean isRecoveryMode) {
     if (isRecoveryMode) {
       return Launcher.RECOVERY;
     }
     return Launcher.NORMAL;
   }
+
```
namtruong: We might want to avoid adding logics into the Launcher as these are hidden and not covered by unit tests

tessera-context/src/main/java/com/quorum/tessera/context/internal/DefaultRuntimeContext.java
```diff
@@ -123,6 +131,16 @@ public boolean isOrionMode() {
     return orionMode;
   }
 
+  @Override
+  public boolean getOutputServerURIs() {
+    return outputServerURIs;
+  }
+
```
namtruong, line 138: This might never be called

namtruong (dismissed):
This can be done within the TesseraCommand rather in Launcher - using the Mixin pattern (similar to `PidFileMixin`)

macfarla (dismissed):
We need this to make testing easy and safe and reliable

Krish1979:
I am not fully convinced about the use case and Tessera is not designed to grab any port available. I discussed this with Tony and he agrees with my comment. Please discuss with Tony before proceeding further work on this change.

macfarla:
> I am not fully convinced about the use case and Tessera is not designed to grab any port available. I discussed this with Tony and he agrees with my comment. Please discuss with Tony before proceeding further work on this change.

@Krish1979 who is Tony?

macfarla:
> I am not fully convinced about the use case and Tessera is not designed to grab any port available.

@Krish1979 Tessera will claim a free port only when the port is not specified in the config

macfarla:
> This can be done within the TesseraCommand rather in Launcher - using the Mixin pattern (similar to `PidFileMixin`)

@mark-terry confirming you have addressed this feedback? @namtruong can you confirm the requested changes were made and dismiss the "changes requested" status?

antonydenyer:
Just wanted to clarify if this was only to be used for testing? If so can we make the `--outputServerURIs` option hidden and also maybe make it 'experimental' ie `--XoutputServerURIs`.

namtruong (dismissed):
If this feature is added just to aid the acceptance testing process (i.e. starting a network of Tessera nodes) - my opinion is it's not a good enough justification. A recommended alternative is try to allocate free ports, generate the relevant Tessera configs then start Tessera nodes with them.

If this is a requested requirement i.e. genuine business use case, have we considered other implications such as : 
- If Tessera does not have a predefined port, what would go into the url of `peers` list ?
- Tessera's partyinfo exchange will not send out its actual p2p URI (I assume the port in config will be 0)
- In case of restart, Tessera might have all of its ports changed. I don't think GoQuorum or Besu at the moment have the capability to cope with this.


Additionally, those points aside, any code changes need tests (unit + integration tests) and they should be written first. The build tollgate won't be passed without 100% coverage.

Krish1979:
I raised the same point with Tony around Tessera restart,, how will Quorum reestablish connection with Tessera after restart if the port is not static,. Currently Quorum reconnects automatically but that may not be the case after this change and would expect Quorum to restart as well.

mark-terry:
I've done some refactoring and added test coverage for the new mixin. I've made the CLI option hidden and added the experimental prefix.

Just to clarify: this PR isn't adding the ability to use dynamic ports by specifying port 0 in the config - that feature already exists. I've just added functionality to record the server URIs in a specified location.

cli/cli-api/src/main/java/com/quorum/tessera/cli/parsers/ServerURIOutputMixin.java
```diff
@@ -0,0 +1,111 @@
...
+      // Add a shutdown hook to clean them up
+      Runtime.getRuntime()
+          .addShutdownHook(
+              new Thread(
+                  () -> {
+                    try {
+                      for (final Path uriFilePath : uriPaths) {
+                        Files.delete(uriFilePath);
+                      }
+                    } catch (final Exception ex) {
```
taccatisid: Can we move the try-catch block inside the for loop so that we continue trying to clean files up even after throwing an exception on one of them?

cli/cli-api/src/main/java/com/quorum/tessera/cli/parsers/ServerURIOutputMixin.java
```diff
@@ -0,0 +1,111 @@
...
+  public void updateConfig(final Path outputServerURIPath, final Config config) {
+    if (outputServerURIPath != null) {
+      config.setOutputServerURIPath(outputServerURIPath);
+    }
+  }
+
+  public static void writeServerURIsToFile(
+      final Path outputServerURIPath, final List<TesseraServer> servers) {
+    try {
+      final TesseraServer q2tServer =
```
taccatisid: Minor style issue: this code is quite repetitive. Couldn't we have a single loop of `servers`, call `writeURIFile` on every single element and just do nothing if `fileNameFromAppType` returns `null`?

I am assuming we do not need the `findFirst()` on the stream to eliminate duplicate instances of the same service.

Krish1979:
> I've done some refactoring and added test coverage for the new mixin. I've made the CLI option hidden and added the experimental prefix.
> 
> Just to clarify: this PR isn't adding the ability to use dynamic ports by specifying port 0 in the config - that feature already exists. I've just added functionality to record the server URIs in a specified location.

The concern was never around this CLI functionality. It is more on the deliberation of unassigned port for Tessera in acceptance tests which will negate some of Tessera "features" i.e., how will a Tessera node discover another node in the network, what will be specified in peers ?? Are there any work around in the acceptance test suites which is not obvious here ?

cli/cli-api/src/main/java/com/quorum/tessera/cli/parsers/ServerURIOutputMixin.java
mark-terry: Good idea.

cli/cli-api/src/main/java/com/quorum/tessera/cli/parsers/ServerURIOutputMixin.java
mark-terry: Good suggestion.

macfarla (approved):
Tests look good. This is a hidden option so won't disrupt any existing users

macfarla:
> If this feature is added just to aid the acceptance testing process (i.e. starting a network of Tessera nodes) - my opinion is it's not a good enough justification. A recommended alternative is try to allocate free ports, generate the relevant Tessera configs then start Tessera nodes with them.
> 
> If this is a requested requirement i.e. genuine business use case, have we considered other implications such as :
> 
> * If Tessera does not have a predefined port, what would go into the url of `peers` list ?
> * Tessera's partyinfo exchange will not send out its actual p2p URI (I assume the port in config will be 0)
> * In case of restart, Tessera might have all of its ports changed. I don't think GoQuorum or Besu at the moment have the capability to cope with this.
> 
> Additionally, those points aside, any code changes need tests (unit + integration tests) and they should be written first. The build tollgate won't be passed without 100% coverage.

^ valid questions @namtruong - however this functionality is not added by this PR. The "use a free port" functionality already exists in Tessera. Maybe we want to doc that it's not recommended for production use, or if you want to do so here are the things you would need to do - but that's separate to this PR. 

Tests have been added so the build tollgate now passes.

namtruong:
> > If this feature is added just to aid the acceptance testing process (i.e. starting a network of Tessera nodes) - my opinion is it's not a good enough justification. A recommended alternative is try to allocate free ports, generate the relevant Tessera configs then start Tessera nodes with them.
> > If this is a requested requirement i.e. genuine business use case, have we considered other implications such as :
> > 
> > * If Tessera does not have a predefined port, what would go into the url of `peers` list ?
> > * Tessera's partyinfo exchange will not send out its actual p2p URI (I assume the port in config will be 0)
> > * In case of restart, Tessera might have all of its ports changed. I don't think GoQuorum or Besu at the moment have the capability to cope with this.
> > 
> > Additionally, those points aside, any code changes need tests (unit + integration tests) and they should be written first. The build tollgate won't be passed without 100% coverage.
> 
> ^ valid questions @namtruong - however this functionality is not added by this PR. The "use a free port" functionality already exists in Tessera. Maybe we want to doc that it's not recommended for production use, or if you want to do so here are the things you would need to do - but that's separate to this PR.
> 
> Tests have been added so the build tollgate now passes.

The functionality does not exist in Tessera and we never promote it. You might be able to start Tessera with port 0 but it will affect other Tessera features and as a result make the whole network not function properly (some examples include partyinfo exchange, influxdb monitoring, HA setup with proxy, and possibly some more if being tested thoroughly which I doubt there has been any).

The concerns are not with the functionality raised by this PR but the justification for its necessity - and the justifications mentioned have raised more problems that have not been thoroughly tested and answered

namtruong (commented):
The conversation on this PR has dragged on for some time now - so will leave it to @Krish1979 and @antonydenyer to make the final decision.

My personal view on this is 

> The use case for this is running the Tessera app without predefined ports in an acceptance test environment.

is not a good use case - and the concerns raised have not been addressed/answered properly.

mark-terry:
I've updated the description to more accurately reflect the PR iterations and clarified the use case.

mark-terry:
Feedback taken onboard. Closing this PR due to a change in approach.

PR #1387, opened 2022-01-27
Add dumb-init to base image to help with term signals

antonydenyer:
This isn't directly a problem for tessera when running a container directly with something like `docker run` however it's fairly common practice to run a bunch of your own entry scripts (as we do in the dev-quickstart). I suspect a lot of people use them as a template. 

The problem with this approach is that SIGTERM signals no longer get passed along to tessera and tessera will be running on a different pid. dumb-init solves this problem.

https://github.com/Yelp/dumb-init

If there is no appetite for adding this to the base image I will update the dev-quickstarts with a custom docker image that will inherit from tessera.

### DIFF
--- a/docker/enclave.Dockerfile
+++ b/docker/enclave.Dockerfile
@@ -15,4 +15,7 @@ FROM adoptopenjdk/openjdk11:alpine
 
 COPY --from=extractor /install/enclave-jaxrs/ /tessera
 
-ENTRYPOINT ["/tessera/bin/enclave-jaxrs"]
+RUN apk add dumb-init
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+CMD ["/tessera/bin/tessera"]
--- a/docker/enclave.Dockerfile
+++ b/docker/enclave.Dockerfile
@@ -18,4 +18,4 @@ COPY --from=extractor /install/enclave-jaxrs/ /tessera
 RUN apk add dumb-init
 
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
-CMD ["/tessera/bin/tessera"]
+CMD ["/tessera/bin/enclave-jaxrs"]
--- a/docker/tessera.Dockerfile
+++ b/docker/tessera.Dockerfile
@@ -15,4 +15,7 @@ FROM adoptopenjdk/openjdk11:alpine
 
 COPY --from=extractor /install/tessera/ /tessera
 
-ENTRYPOINT ["/tessera/bin/tessera"]
+RUN apk add dumb-init
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+CMD ["/tessera/bin/tessera"]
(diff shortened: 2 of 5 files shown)


build.gradle
```diff
@@ -116,7 +116,7 @@ allprojects {
         implementation "org.bouncycastle:bcpkix-jdk15on:1.68"
         implementation "org.bouncycastle:bcprov-jdk15on:1.68"
 
-        implementation "com.h2database:h2:2.0.206"
+        implementation "com.h2database:h2:2.1.210"
```
chris-j-h: Dupe of #1388 change - i'd say remove from here so we can track the change easier in the standalone PR - just in case this gets merged in before

docker/enclave.Dockerfile
```diff
@@ -15,4 +15,7 @@ FROM adoptopenjdk/openjdk11:alpine
 
 COPY --from=extractor /install/enclave-jaxrs/ /tessera
 
-ENTRYPOINT ["/tessera/bin/enclave-jaxrs"]
\ No newline at end of file
+RUN apk add dumb-init
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+CMD ["/tessera/bin/tessera"]
```
chris-j-h: ```CMD ["/tessera/bin/enclave-jaxrs"]```

chris-j-h (changes requested):
Couple of comments.

https://github.com/Yelp/dumb-init#using-a-shell-for-pre-start-hooks - Does this mean there could be impacts for any users that are using custom CMDs to do some additional work?

What are the size differences by adding dumb-init to the images?  

Can you give a bit more detail on what dev-quickstart is doing in its entryscripts and what problems this causes without dumb-init.  It would help understand the change and may help others who are searching for solutions if they have the same problem.

antonydenyer:
> Couple of comments.
> 
> https://github.com/Yelp/dumb-init#using-a-shell-for-pre-start-hooks - Does this mean there could be impacts for any users that are using custom CMDs to do some additional work?

Yes, they would need to change the cmd or the entry point. If you had something like:
```
FROM quorumengineering/tessera:latest
COPY tessera-config.json
CMD ["-configfile", /tessera-config.json"]
```
you would need to change the Dockerfile to be
```
FROM quorumengineering/tessera:latest
COPY tessera-config.json
CMD ["/tessera/bin/tessera", "-configfile", /tessera-config.json"]
```
or 

```
FROM quorumengineering/tessera:latest
COPY tessera-config.json
CMD ["-configfile", /tessera-config.json"]
ENTRYPOINT ["/tessera/bin/tessera"]
```

> 
> What are the size differences by adding dumb-init to the images?
> 

less than a kb

> Can you give a bit more detail on what dev-quickstart is doing in its entryscripts and what problems this causes without dumb-init. It would help understand the change and may help others who are searching for solutions if they have the same problem.

It's a combination of problems. In the quick start we're changing the config based on env vars and then starting tessera. The problem with this is that the main entry point becomes the bash script and tessera ends up starting on a non 1 PID. 

The main problem with this is that tessera will not react to a sigterm, so you could have strange things like vertx not finishing a response or worse not closing a db cleanly, partial commits that kind of thing. 

To be clear, this is not a problem with tessera per se and thinking about it more maybe the right answer is to fix the dev-quick-start to follow better practices.

antonydenyer:
Update the code in https://github.com/ConsenSys/quorum-dev-quickstart/pull/136 to contain better examples

PR #1390, opened 2022-01-31
[CI] Trigger releases from repo page, fix PR builds from forks

chris-j-h:
1. Fix PR builds from forks 
    * Don't use secret for image name 
    * Fork PRs don't have access to secrets so the builds for these PRs was failing when trying to create a local image for use in the atests.  The image name is publicly known so does not need to be stored as a secret.

1. Set up workflow_dispatch for releases which can be triggered from github.com
    * workflow_dispatch allows the workflow to be run from branches other than master - making future patch releases much simpler.  
    * To prevent accidental release of the wrong branch, require users to provide the version number which is checked against version.txt.

1. Upgrade dependencycheck gradle plugin

1. Cleanup unused slack notification steps


chris-j-h:
Closing to raise from fork

PR #1395, opened 2022-02-09
P2P Messaging Code Changes

shashik90:
The PR is from initial review that @sjh-mclabs raised. We have added more commit to that such as "Remove All references of SHA1 and Truncate SHA256 to 20 Bytes" and "unit test code coverage". Kindly review it.


namtruong:
Please provide more context for this change

shashik90:
> Please provide more context for this change

We are extending Tessera to support private-slash-peer-to-peer-messaging. 
Using Tessera for (asynchronous) messaging in this way would be structurally similar to how it is used for propagating private transactions: a client seeking to send a message to some other entity would make a request to an endpoint on its Tessera instance, specifying the message along with the public key of the recipient. The Tessera Enclave would generate a (shared) key between the private key corresponding to the sender's public key and the recipient's public key and use the shared key to encrypt the message contents which would be forwarded on to the recipient's Tessera instance which can retrieve and decrypt the message, on demand, for the recipient.
Key differences with how private transactions are handled by Tessera:
Messages would only be persisted on the receiving side
Messages are not intended to be long-lived; clients are expected to delete messages once they've been read/processed
Barring some other mechanism for notifying clients that a new message has arrived, requiring their attention, clients would need to poll for new messages (and to delete read messages to make polling less inefficient)

@pinges   As you have previously reviewed this PR with @viathefalcon and suggested the changes to him, please feel free to review the PR.  @viathefalcon left the organisation hence we had to create a new PR which includes the changes reviewed earlier as well as changes you suggested.

namtruong:
Thanks for the response. 

Could you please elaborate the use-case/requirements for private-slash-peer-to-peer-messaging? What is sending those messages and what are they used for ?

The primary purpose of Tessera is to act as a privacy manager for privacy-enabled Ethereum clients such as GoQuorum and Besu. It's not likely to be within the application's scope to support external clients or any objectives other than handling private transactions.

Perhaps @pinges can shed some more lights on this ?

macfarla:
> [CLA assistant check](https://cla-assistant.io/ConsenSys/tessera?pullRequest=1395) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/ConsenSys/tessera?pullRequest=1395) before we can accept your contribution.**2** out of **5** committers have signed the CLA.✅ Chirag-786✅ shashik90❌ Admin❌ gsmorag❌ sjh-mclabs
> 
> **Admin** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/ConsenSys/tessera?pullRequest=1395) it.

Hi @shashik90 - there is one commit from `Admin` which I'm guessing doesn't resolve to a real user. I think you're going to have to manually fix up that commit and redo it with a real user. Same for the commits by gsmorag and sjh-mclabs if they aren't available to sign the CLA. I think your best bet is to cherry-pick the valid commits to a new branch, copy-paste the other commits in as needed (recommend doing them in the same order as the original to avoid conflicts).

chris-j-h:
@shashik90 @macfarla you may be able to do something like `git commit --amend --author`

> 
> Hi @shashik90 - there is one commit from `Admin` which I'm guessing doesn't resolve to a real user. I think you're going to have to manually fix up that commit and redo it with a real user. Same for the commits by gsmorag and sjh-mclabs if they aren't available to sign the CLA. I think your best bet is to cherry-pick the valid commits to a new branch, copy-paste the other commits in as needed (recommend doing them in the same order as the original to avoid conflicts).

tessera-jaxrs/thirdparty-jaxrs/src/main/java/com/quorum/tessera/thirdparty/messaging/SendMessageRequest.java
```diff
@@ -0,0 +1,48 @@
...
+public class SendMessageRequest {
+
+  @Schema(
+      description = "Public key identifying the message sender",
+      type = "string",
+      format = "base64")
+  private String from;
+
+  @Schema(
+      description = "Public key identifying the message sender",
```
pinges, line 14: message receipient?

PR #1401, opened 2022-02-22
Adding delete All method for deleting based on public key

PuneetGoel80:


macfarla:
Hi @PuneetGoel80 what is the use case for this change?

PR #1410, opened 2022-03-31
update jackson version

macfarla:
and allow for different version for databind

removed this plugin "io.swagger.core.v3.swagger-gradle-plugin"
it pulls in an earlier version of jackson-databind and doesn't respect the constraint
Maybe there's a simpler way but I couldn't find it.

This means we would need to use a different method to generate the swagger docs. eg Teku has a CLI command https://github.com/ConsenSys/teku/blob/0a6ba3b63d9f59b8fc5167cfbee7d5402ab29a84/teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugToolsCommand.java

Also added Ignore for tests of the OpenApi generation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -160,6 +160,11 @@ allprojects {
         testImplementation "org.apache.tuweni:tuweni-rlp:2.0.0"
         testImplementation "org.apache.tuweni:tuweni-bytes:2.0.0"
 
+        constraints {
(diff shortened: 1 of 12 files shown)


macfarla:
the problem is this plugin id "io.swagger.core.v3.swagger-gradle-plugin"

it pulls in an earlier version of jackson-databind which doesn't respect the constraint

namtruong (approved):

PR #1443, opened 2022-04-08
docker image: change openjdk11 with alpine to debianslim with arm64 support

baptiste-b-pegasys:
## PR Description

In order to use tessera on the mac m1 (with the arm processor), the docker image has to support the arm64 architecture.

The current openjdk11:alpine doesn't support arm64, but the openjdk11:debianslim does.

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog


Krish1979 (approved):

macfarla (approved):

chris-j-h (changes requested):
Using this base considerably increases the size of the image

```
quorumengineering/quorum   debianslim           29aeeb54d1ba   33 seconds ago   468MB
quorumengineering/quorum   latest               79024fb87590   12 days ago      88.5MB
```

Perhaps it would be better to create a separate m1 image if this is the only option, instead of incurring a large storage hit on all other users?

Plus, what about the enclave and vault images we provide?

baptiste-b-pegasys:
88MB for a JVM is pretty low, I think you are comparing goquorum with tessera

`adoptopenjdk/openjdk11       alpine        16e2d39c5bca   45 hours ago    342MB`

chris-j-h (approved):

chris-j-h:
> 88MB for a JVM is pretty low, I think you are comparing goquorum with tessera
> 
> `adoptopenjdk/openjdk11 alpine 16e2d39c5bca 45 hours ago 342MB`

right you are - typo when building the image :)

PR #1448, opened 2022-04-28
Add wget and netcat into base repo

antonydenyer:
## PR Description
Wget and netcat were available in the previous base image and commonly used for upcheck


achraf17 (approved):
LGTM

PR #1457, opened 2022-05-26
set SNI flags to false

macfarla:

Set SNI flags to false. Default behaviour changed in jetty upgrade so this brings back the previous behaviour

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog


Krish1979 (approved):

Krish1979 (changes requested):
I don't think we should disable the FLAG. The correct solution would be to recreate correct certificates to match the SNI of the host

PR #1464, opened 2022-06-22
Improve keygen CLI output, fix security advisories in key vault modules

chris-j-h:
* Print keygen results to stdout as current results are buried in log msgs
    * Output includes: generated public key, key-storage specific information (e.g. filepath, vault secret id)
* Update key-vault modules to address non-critical security advisories

Example keygen command and output:

```
$ tessera keygen --keyout /keys/test

1 keypair(s) generated:
	1: type=file, pub=WO5Wj+tXIxf8Amrv6RiOuAqXWGESatnscjal8AmV1AY=
		pub: path=/keys/test.pub
		prv: path=/keys/test.key
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 import java.time.Duration
 
 plugins {
-  id "org.owasp.dependencycheck" version "7.0.4.1"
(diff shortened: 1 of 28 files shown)


antonydenyer (approved):
Nice

PR #1489, opened 2022-10-03
Possibly mitigate issues with PKIX path building failed

antonydenyer:
When making requests in CompletableFuture the client occasionally 'loses' the security context. Possibly because it's not thread safe in either jetty
or sslcontext.

## Fixed Issue(s)
https://github.com/ConsenSys/tessera/issues/1468

## Changelog


pinges (approved):
LGTM

